PR #9, opened 2019-03-20
Api integration

lboeman:
Additions in this PR:
- Uses API as the data source for all metadata displays.
- Forms Post data to the API, display errors to the user on failure, redirect to associated metadata object on success.
- Some code refactoring to remove hard coded links. 
- Adds some abstraction. I made a conscious effort here to repeat code blocks so that needed refactoring will be more apparent down the road. I didn't want to commit to too many abstractions early on while the data model is still changing.

This code depends on the changes in introduced in: https://github.com/SolarArbiter/solarforecastarbiter-api/pull/51


wholmgren:
are you going to make static-design the master branch after this?

lboeman:
> are you going to make static-design the master branch after this?

Yes, I think merging this back into master is a move in the right direction. I don't think we'll have a functional dev-api spun up until we get the auth and database sorted out, which would be my only concern about having it in the master branch.

Do you want to review this before it is merged? Now that the bulk of the conversion is done, I can work in smaller PRs.

sfa_dash/blueprints/form.py
```diff
@@ -1,42 +1,277 @@
...
+        return f'{hours:02d}:{minutes:02d}'
+
+    def parse_timedelta(self, data_dict, key_root):
+        value = int(data_dict[f'{key_root}_number'])
+        units = data_dict[f'{key_root}_units']
+        if units == 'minutes':
+            return value
+        if units == 'hours':
+            return value * 60
+        else:
```
wholmgren: are we allowing seconds? 

In any case, perhaps the `else` block should raise a `ValueError('invalid units')` instead of defaulting to something.

sfa_dash/blueprints/form.py
```diff
@@ -1,42 +1,277 @@
...
+        Returns
+        -------
+        dictionary
+            Form data formatted to the API spec.
+        """
+        modeling_keys = ['ac_capacity', 'dc_capacity',
+                         'temperature_coefficient', 'axis_azimuth',
+                         'tracking_type', 'backtrack',
+                         'axis_tilt', 'ground_coverage_ratio',
+                         'surface_tilt', 'surface_azimuth']
```
wholmgren: we added `dc_losses` and `ac_losses` to the data model.

sfa_dash/templates/data/metadata/site_metadata.html
```diff
@@ -4,26 +4,28 @@
...
   {% if modeling_parameters %}
     <li><b>Modeling Parameters:</b>
       <div class="data-metadata-field-modeling-parameters pl-2">
         <ul class="data-metadata-fields">
-        {{ macro.li('AC Power', modeling_parameters['ac_power'] | string + ' MW') }}
-        {{ macro.li('DC Power', modeling_parameters['dc_power'] | string + ' MW') }}
-        {{ macro.li('Temperatue Coefficient', modeling_parameters['temperature_coefficient'] | string +' 1/C')}}
+        {{ macro.li('AC Capacity', modeling_parameters['ac_capacity'] | string, 'MW') }}
+        {{ macro.li('DC Capacity', modeling_parameters['dc_capacity'] | string, 'MW') }}
+        {{ macro.li('Temperatue Coefficient', modeling_parameters['temperature_coefficient'] | string, '1/C')}}
```
wholmgren: losses here too

sfa_dash/templates/forms/forecast_data_form.html
```diff
@@ -0,0 +1,49 @@
...
+<form action="{{ url_for('forms.upload_forecast_data', uuid=uuid) }}" method="post" id="forecast-upload-form" enctype="multipart/form-data">
+  <div class="form-group">
+	<div class="form-element">
+	  <label>My data is formatted in:</label><br>
+	  <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="csv" checked/>CSV
+      <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="json"/>JSON
+	</div>
+	<div class="form-element full-width">
+      <div class="upload-example csv show">
+		<p>Observation data in CSV format should follow the formatting of the example below.</p>
```
wholmgren: Forecast, not obs

sfa_dash/templates/forms/forecast_data_form.html
```diff
@@ -0,0 +1,49 @@
...
+	    <pre class="example-text">
+# optional header, ignored by Solar Forecast Arbiter 
+timestamp,value
+2018-11-22T12:01:48Z,10.23 
+2018-11-22T12:07:38Z,10.67</pre>
+	  </div>
+	  <div class="upload-example json collapse">
+		<p>Observation data in JSON format should follow the formatting of the example below.</p>
```
wholmgren: same

sfa_dash/templates/forms/forecast_data_form.html
```diff
@@ -0,0 +1,49 @@
...
+	    <pre class="example-text">{ 
+    “id”: “testid”, # optional, for uploader’s record 
+    “values”: [ 
+        { 
+           “timestamp”: “2018-11-22T12:01:48Z”, 
+            “value”: 10.23,
```
wholmgren: indentation off here and below

sfa_dash/blueprints/form.py
lboeman: I can allow seconds here if need be. I will have to allow for floats in the API as well.

sfa_dash/blueprints/form.py
lboeman: :+1: I'll add this to the API as well.

sfa_dash/blueprints/form.py
wholmgren: wait, what is this else case supposed to handle?

lboeman: It handles days. I'm going to update this to return 'NaT' as default so that the API returns an error with any others found in the form.

wholmgren: I'm totally lost then. Maybe a doc string would help. Maybe it's supposed to be 1440 instead of 3600? Also use `if... elif... elif... else: NaT` to put it all into a single block.

lboeman: You're right. I updated this and the API to use minutes over seconds and messed up the values in both places. Thanks for catching it. 
Do you think there is a case where we need to allow for defining timedeltas in seconds, and therefore allowing the API to store them as floats with units of minutes?

wholmgren: no, I don't think we need to support timedeltas of seconds. my comment about seconds was only because 3600 is the number of seconds in an hour, so I thought you might be handling seconds.

lboeman:
Alright, I fixed the timedelta parsing with if/elif/else for minutes/hours/days and factors of 1/60/1440 respectively with a default of 'NaT'. Added ac/dc loss factor to the Site form and metadata display. Fixed the leftover Observation labels in the forecast upload templates, and corrected the indentation.

PR #11, opened 2019-03-21
Download forms

lboeman:
Adding Data download forms for observations and forecasts.


lboeman:
I think this is ready for review. the bulk of the updates are in sfa_dash/blueprints/form.py fx and obs data_download templates. We don't have a simple way of determining available data at the moment, so I commented out their placeholders in the templates. When a good way to determine available data is introduced, The start/end widgets should have defaults. Perhaps the most recent day of data?

sfa_dash/blueprints/form.py
```diff
@@ -280,6 +282,78 @@ def post(self, uuid):
...
 
 
+class DownloadForm(BaseView):
+    def __init__(self, data_type):
+        self.data_type = data_type
+        if data_type == 'observation':
+            self.template = 'forms/obs_data_download_form.html'
+            self.metadata_template = 'data/metadata/observation_metadata.html'
+            self.api_handle = observations
+        if data_type == 'forecast':
```
wholmgren:
```suggestion
        elif data_type == 'forecast':
```

sfa_dash/blueprints/form.py
```diff
@@ -280,6 +282,78 @@ def post(self, uuid):
...
+            self.template = 'forms/fx_data_download_form.html'
+            self.metadata_template = 'data/metadata/forecast_metadata.html'
+            self.api_handle = forecasts
+
+    def format_params(self, form_data):
+        """Parses start and end time and the format from the posted form
+        and returns headers and query parameters for requesting the data.
```
wholmgren: confusing. maybe

        """Parses start time, end time, and the format from the posted form.
        Returns headers and query parameters for requesting the data.

sfa_dash/blueprints/form.py
```diff
@@ -280,6 +282,78 @@ def post(self, uuid):
...
+                                                  params=params)
+        if data_request.status_code != 200:
+            abort(404)
+        if form_data['format'] == 'application/json':
+            data = data_request.json()
+            response = make_response(json.dumps(data))
+            response.headers.set('Content-Type', 'application/json')
+            response.headers.set(
+                'Content-Disposition', 'attachment', filename='data.json')
+        if form_data['format'] == 'text/csv':
```
wholmgren:
```suggestion
        elif form_data['format'] == 'text/csv':
```

sfa_dash/blueprints/form.py
```diff
@@ -280,6 +282,78 @@ def post(self, uuid):
...
+            csv_data = data_request.text
+            response = make_response(csv_data)
+            response.headers.set('Content-Type', 'text/csv')
+            response.headers.set(
+                'Content-Disposition', 'attachment', filename='data.csv')
```
wholmgren: else: raise ValueError('invalid format')

or similar?

sfa_dash/blueprints/form.py
```diff
@@ -280,6 +282,78 @@ def post(self, uuid):
...
+    def __init__(self, data_type):
+        self.data_type = data_type
+        if data_type == 'observation':
+            self.template = 'forms/obs_data_download_form.html'
+            self.metadata_template = 'data/metadata/observation_metadata.html'
+            self.api_handle = observations
+        if data_type == 'forecast':
+            self.template = 'forms/fx_data_download_form.html'
+            self.metadata_template = 'data/metadata/forecast_metadata.html'
+            self.api_handle = forecasts
```
wholmgren: these are mutually exclusive and should probably error if data_type is anything else.

sfa_dash/templates/forms/fx_data_download_form.html
```diff
@@ -0,0 +1,70 @@
...
+	  <div class="download-example json collapse">
+	    <p>JSON downloads will have the following format.</p>
+		<pre class="example-text">{
+    “id”: “f79e4f84-e2c3-11e8-9f32-f2801f1b9fd1',
+	"_links": {
+		"metadata": "https://api.solarforecastarbiter.org/forecasts/f79e4f84-e2c3-11e8-9f32-f2801f1b9fd1/metadata",
+	},
+    “values”: [
+        { 
+           “timestamp”: “2018-11-22T12:01:48Z”,
```
wholmgren: make the timestamps consistent with csv. add the 3rd point here too

sfa_dash/templates/forms/fx_data_download_form.html
```diff
@@ -0,0 +1,70 @@
...
+	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
+	</div>
+	<div class="form-element full-width">
+	  <div class="download-example csv show">
+	  <p>CSV downloads will have the following format.</p>
+	  <pre class="example-text">
+# name: DA Power
+# forecast_id: f79e4f84-e2c3-11e8-9f32-f2801f1b9fd1
+timestamp,value
+2018-11-22T12:01:00Z,10.23
```
wholmgren: probably no place in the US that would have positive power at this time. use 18 instead of 12 in all examples.

wholmgren: I know it's dumb but it's one less thing for people to be caught up on.

sfa_dash/templates/forms/fx_data_download_form.html
```diff
@@ -0,0 +1,70 @@
...
+	  <label for="format">Format</label><br/>
+	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
+	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
+	</div>
+	<div class="form-element full-width">
+	  <div class="download-example csv show">
+	  <p>CSV downloads will have the following format.</p>
+	  <pre class="example-text">
+# name: DA Power
+# forecast_id: f79e4f84-e2c3-11e8-9f32-f2801f1b9fd1
```
wholmgren: route is `forecasts` with an s. should it be consistent?

sfa_dash/templates/forms/obs_data_download_form.html
```diff
@@ -0,0 +1,66 @@
...
+      <label for="format">Format</label><br/>
+      <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
+      <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
+    </div>
+    <div class="form-element full-width">
+      <div class="download-example csv show">
+      <p>CSV downloads will have the following format.</p>
+      <pre class="example-text">
+# name: GHI Instrument 1
+# obs_id: 123e4567-e89b-12d3-a456-426655440000
```
wholmgren: definitely not `obs_id`. observations_id? id?

sfa_dash/templates/forms/fx_data_download_form.html
wholmgren: or just `id` for consistency with the json format

sfa_dash/templates/forms/obs_data_download_form.html
```diff
@@ -0,0 +1,66 @@
...
+timestamp,value,quality_flag
+2018-11-22T12:01:00Z,10.23,0
+2018-11-22T12:02:00Z,10.67,0
+2018-11-22T12:03:00Z,10.67,0
```
wholmgren: same comments about csv and json times as above.

sfa_dash/templates/forms/obs_data_download_form.html
lboeman: I agree. I think this should be a sweeping change though, maybe made in it's own PR to each of our repos.

sfa_dash/templates/forms/fx_data_download_form.html
lboeman: I think forecast_id, observation_id and site_id make more sense as the id labels for single objects. The json should show forecast_id as well.
I'll have to fix the discrepancies between the API and data model again, as the json endpoints don't include name or links, and the csv endpoint does not write metadata as a header comment.

lboeman:
After thinking about the form validation errors, I think you were right. I opted to raise an exception when an invalid selection is received from the form because there are only two cases where that's possible:
- a developer added new options to the select list in the HTML and didn't add them to the code, throwing an error seems the fastest way to alert them to that.
- html was modified in the browser to allow for a selection that we did not define.
in both cases, I think the flask app should fail before sending an API request.

PR #14, opened 2019-04-03
authentication via Auth0

alorenzo175:
uses flask-dance blueprint to add a /login/auth0 path to log users in. Right now, tokens are stored in the client session cookie which ok for testing for now. Should implement storage on the backend and then we can also use refresh tokens. Also need the client id and secret from auth0 even for local testing

### DIFF
--- a/sfa_dash/api_interface/__init__.py
+++ b/sfa_dash/api_interface/__init__.py
@@ -1,7 +1,9 @@
 """Helper functions for all Solar Forecast Arbiter /sites/* endpoints.
 """
 from flask import current_app as app
-import requests
+
+
+from sfa_dash import auth0
 
 
 def get_request(path, **kwargs):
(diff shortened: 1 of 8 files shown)


sfa_dash/api_interface/__init__.py
```diff
@@ -17,7 +19,13 @@ def get_request(path, **kwargs):
...
         The api response.
     """
-    return requests.get(f'{app.config["SFA_API_URL"]}{path}', **kwargs)
+    try:
+        session = auth0  # already has token ready to use for api
+    except Exception:
+        # make a requests session?
+        # redirect to login?
+        raise
+    return session.get(f'{app.config["SFA_API_URL"]}{path}', **kwargs)
```
lboeman: I think that we don't necessarily need to make this try/except here if all the endpoints besides / are protected, we can probably safely assume that the user at least has a session.

sfa_dash/blueprints/auth0.py
```diff
@@ -0,0 +1,63 @@
...
+    from flask import _app_ctx_stack as stack
+except ImportError:
+    from flask import _request_ctx_stack as stack
+from flask import url_for, redirect, current_app, session
+from flask.globals import LocalProxy, _lookup_app_object
+from flask_dance.consumer import OAuth2ConsumerBlueprint
+from six.moves.urllib.parse import urlencode
+
+
+auth0 = LocalProxy(partial(_lookup_app_object, 'auth0_oauth'))
```
lboeman: Although this is how the flask dance examples do it, naming the session for the Oauth provider was kind of confusing here, and it even gets renamed to session in the api interface. Maybe just call it auth0_session or something similar?

PR #26, opened 2019-04-05
catch no refresh token errors

lboeman:
Should log users out when an error is thrown due to missing refresh tokens. Haven't triggered this in the way users will yet, because the flask session has been reset as I've been working, so we should hold off on merging until it is adequately tested.


lboeman:
When manually setting the token's expired-at attribute, the InvalidClientIdError is raised and caught and the user is redirected to login, but I haven't been able to mock the token properly to get around the `protect_endpoint` wrapper. 

I think ideally, the way to test this is to mock the Oauth2Session's load_token() method, so that it's agnostic of the storage class we end up using, but I'm not sure how to mock the `oauth_request_session` Proxy setup in the `auth0` endpoint which provides a handle to the session elsewhere..
@alorenzo175 do you have any ideas? Or is this even worth dealing with until we have sql storage?

alorenzo175:
hmm, does not look easy. Maybe if you define the session storage in ``sfa_dash/__init__.py``and pass that to make_auth0_blueprint, you can mock the storage object to return the expired token. And yes, probably worth figuring out how to make this work even without sql

lboeman:
These tests are working now when provided the correct Auth0 Configuration. We'll have to figure out how to either provide legitimate client secrets to the integration tests, or provide some test configuration to hit a dummy Auth0 app. But perhaps there should be a PR dedicated to getting integration tests and testing auth setup.

PR #28, opened 2019-04-05
CDF Download and Upload forms

lboeman:


sfa_dash/templates/forms/cdf_forecast_upload_form.html
```diff
@@ -0,0 +1,48 @@
...
+      <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="json"/>JSON
+	</div>
+	<div class="form-element full-width">
+      <div class="upload-example csv show">
+		<p>Forecast data in CSV format should follow the formatting of the example below.</p>
+	    <pre class="example-text">
+# optional header, ignored by Solar Forecast Arbiter 
+timestamp,value
+2018-11-22T12:01:48Z,10.23 
+2018-11-22T12:07:38Z,10.67</pre>
```
wholmgren: Where is this copy pasta coming from? Forecasts and observations should both be at rounded times. So 12:00:00 and 12:05:00 in this example.

PR #32, opened 2019-04-09
manually activated tooltips

### LINKED ISSUE #25: tooltips cover up important info
Some of the tooltips cover up too much other stuff. The main thing is that some of them cover up the name of the field that you're currently filling out. I also worry that they cover up fields that you might want to reference when filling in the active field. 

Maybe a (❔) or similar character next to each field name could be clicked or hovered over instead?

lboeman:
closes #25 
Removes the automatic focus-triggered tooltips and adds question mark button triggers as shown below: 
manual_tooltips

### DIFF
--- a/sfa_dash/blueprints/form.py
+++ b/sfa_dash/blueprints/form.py
@@ -348,9 +348,8 @@ class DownloadForm(BaseView):
             Query parameters start and end, both formatted in iso8601 and
             localized to the provided timezone.
         """
-        timezone = form_data['timezone']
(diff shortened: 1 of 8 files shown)


lboeman:
@wholmgren is this what you had in mind for tooltips(header comment has a screenshot)? I can show it to you live if that's easier to review.

lboeman:
> Yes. Do you think it works better this way or am I causing more/different problems?

I think this is an improvement. I think after becoming familiar, the automatic tooltips may have become frustrating, and they would sometimes have issues with a browser's autofill.

PR #38, opened 2019-04-10
Tooltip display OS compatibility

### LINKED ISSUE #36: form layout issues
Layout issues with Sites, Observations, and Forecast creation forms pictured below. Firefox 66.0.2. Playing with browser width does not solve the problem.

Screen Shot 2019-04-10 at 2 20 22 PM
Screen Shot 2019-04-10 at 2 20 32 PM
Screen Shot 2019-04-10 at 2 20 47 PM

lboeman:
Intends to close #36 
I wasn't able to replicate the issue exactly, but the changes in this pull request brought the behavior and appearance of the forms inline across browsers.

also fixes a couple other issues found on 
- Aligns tooltips with the center of the input element. 
- Explicity set `display` property to inline-block for nested input elements.
- add help (?) button to constant_values and backtrack fields.

### DIFF
--- a/sfa_dash/static/css/styles.css
+++ b/sfa_dash/static/css/styles.css
@@ -233,5 +233,5 @@ a.help-button{
 }
 .input-wrapper{
     width: calc(100% - 29px);
-    display: inline-block;
+    display: inline-flex;
 }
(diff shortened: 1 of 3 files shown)


wholmgren:
Re the title.... Maybe more about mac compatibility than firefox because we saw that chrome and safari on mac had similar issues (safari less so than the others for some reason).

lboeman:
> Re the title.... Maybe more about mac compatibility than firefox because we saw that chrome and safari on mac had similar issues (safari less so than the others for some reason).

These issues were showing up on windows as well. Will fix.

PR #39, opened 2019-04-11
Delete

### LINKED ISSUE #19: Add Delete button

lboeman:
closes #19 
This PR adds a delete button on metadata pages and a confirmation page warning the user that they cannot undo a deletion. 
The post endpoint accepts the delete request and either redirects to the appropriate listing page on success, or redirects to the confirmation with a list of errors.

### DIFF
--- a/sfa_dash/blueprints/delete.py
+++ b/sfa_dash/blueprints/delete.py
@@ -1,0 +1,52 @@
+from flask import url_for, render_template, abort, redirect
+
+from sfa_dash.api_interface import sites, observations, forecasts, cdf_forecast_groups
+from sfa_dash.blueprints.base import BaseView
+from sfa_dash.blueprints.dash import DataDashView
+
+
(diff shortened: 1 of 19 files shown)


sfa_dash/blueprints/delete.py
```diff
@@ -0,0 +1,72 @@
...
+            # the confirmation page, redirect to confirm.
+            return redirect(confirmation_url)
+        delete_request = self.api_handle.delete(uuid)
+        if delete_request.status_code == 400:
+            # Redirect and display errors if the delete request
+            # failed
+            response_json = delete_request.json()
+            errors = response_json['errors']
+            return self.get(uuid, errors=errors)
+        elif delete_request.status_code == 404:
```
alorenzo175: what if the api returns a different error like 500? seems like the user would think the deletion was successful. Should probably catch any other status codes > 400 and raise/handle them

PR #41, opened 2019-04-12
csrf protection

### LINKED ISSUE #34: Add Flask SeaSurf
Protects from csrf attacks by populating a hidden field on forms with a csrf token that must be present for a request to be valid.

lboeman:
Initializes Flask SeaSurf and adds a csrf token to all forms. Most of the changes here are form refactoring. I've extracted most of the repeated html into files in `/templates/form/base`. These templates include the <form> elements, and common input elements such as the hidden csrf token fields.
The files changed in `template/forms` are largely the same, with the code that is now found in the base forms removed.
I've also renamed some of the views with a predictable pattern so that they may be selected programatically by i.e. something like download_{data_type}_data, where data type is 'observation', 'forecast' or 'cdf_forecast'.
closes #34

### DIFF
--- a/sfa_dash/templates/forms/cdf_forecast_download_form.html
+++ b/sfa_dash/templates/forms/cdf_forecast_download_form.html
@@ -1,43 +1,39 @@
-{% extends "dash/data.html" %}
-{# {% set sidebar = True %} #}
+{% extends "forms/base/download_form.html" %}
 {% set page_title = 'Download Forecast data' %}
-{% block content %}
-{{ metadata | safe }}
-<h3>Download data for Forecast 'DA Power'</h3>
-{#<p><b>Available Date Range:</b> <i>2016-01-01T12:00Z</i> to <i>2019-01-01T12:00Z</i></p>#}
-<form action="{{ url_for('forms.download_cdf_forecast_data', uuid=uuid) }}" method="post" id="forecast-download">
-  <div class="form-group">
-	<div class="form-element">
-	  <label for="start">Start time:</label>
-	  <input type="datetime-local" id="start" name="period-start">
-    </div>
-    <div class="form-element">
-	  <label for="end">End time:</label>
-	  <input type="datetime-local" id="end" name="period-end">
-    </div>
-	<div class="form-element full-width">
-	  <label for="format">Format</label><br/>
-	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
-	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
-	</div>
-	<div class="form-element full-width">
-	  <div class="download-example csv show">
-	  <p>CSV downloads will have the following format.</p>
-	  <pre class="example-text">
+{% set form_title = "Download data for Forecast 'DA Power'" %}
+{% set data_type = 'cdf_forecast' %}
+{% block fields %}
+<div class="form-element">
+  <label for="start">Start time:</label>
+  <input type="datetime-local" id="start" name="period-start">
+</div>
+<div class="form-element">
+  <label for="end">End time:</label>
+  <input type="datetime-local" id="end" name="period-end">
+</div>
+<div class="form-element full-width">
+  <label for="format">Format</label><br/>
+  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
+  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
+</div>
+<div class="form-element full-width">
+  <div class="download-example csv show">
+  <p>CSV downloads will have the following format.</p>
+  <pre class="example-text">
 # forecast_id: f79e4f84-e2c3-11e8-9f32-f2801f1b9fd1
 # metadata: https://api.solarforecastarbiter.org/forecasts/f79e4f84-e2c3-11e8-9f32-f2801f1b9fd1/metadata
 timestamp,value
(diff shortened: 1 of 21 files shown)


sfa_dash/templates/forms/cdf_forecast_download_form.html
```diff
@@ -1,43 +1,39 @@
...
-	<div class="form-element full-width">
-	  <label for="format">Format</label><br/>
-	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
-	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
-	</div>
-	<div class="form-element full-width">
-	  <div class="download-example csv show">
-	  <p>CSV downloads will have the following format.</p>
-	  <pre class="example-text">
+{% set form_title = "Download data for Forecast 'DA Power'" %}
```
alorenzo175: This should be dynamic based on the metadata, no?

sfa_dash/templates/forms/cdf_forecast_upload_form.html
```diff
@@ -1,11 +1,8 @@
-{% extends "dash/data.html" %}
+{% extends "forms/base/upload_form.html" %}
 {% set page_title = 'Upload Forecast data' %}
-{% block content %}
-{% include "sections/notifications.html" %}
-{{ metadata | safe }}
-<h3>Upload data for Forecast 'AC Power'</h3>
-<form action="{{ url_for('forms.upload_cdf_forecast_data', uuid=uuid) }}" method="post" id="cdf-forecast-upload-form" enctype="multipart/form-data">
-  <div class="form-group">
+{% set form_title = "Upload data for Forecast 'AC Power'" %}
```
alorenzo175: and here

sfa_dash/templates/forms/forecast_download_form.html
```diff
@@ -1,43 +1,39 @@
...
-	<div class="form-element full-width">
-	  <label for="format">Format</label><br/>
-	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
-	  <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
-	</div>
-	<div class="form-element full-width">
-	  <div class="download-example csv show">
-	  <p>CSV downloads will have the following format.</p>
-	  <pre class="example-text">
+{% set form_title = "Download data for Forecast 'DA Power'" %}
```
alorenzo175: and here

sfa_dash/templates/forms/forecast_upload_form.html
```diff
@@ -1,28 +1,25 @@
...
-	<div class="form-element">
-	  <label>My data is formatted in:</label><br>
-	  <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="csv" checked/>CSV
-      <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="json"/>JSON
-	</div>
-	<div class="form-element full-width">
-      <div class="upload-example csv show">
-		<p>Forecast data in CSV format should follow the formatting of the example below.</p>
-	    <pre class="example-text">
+{% set form_title = "Upload data for Forecast 'AC Power'" %}
```
alorenzo175: an dhere

sfa_dash/templates/forms/observation_download_form.html
```diff
@@ -1,41 +1,37 @@
...
-    <div class="form-element full-width">
-      <label for="format">Format</label><br/>
-      <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="text/csv" checked>CSV</input>
-      <input type="radio" data-toggle="collapse" data-target=".download-example" name="format" value="application/json">JSON</input>
-    </div>
-    <div class="form-element full-width">
-      <div class="download-example csv show">
-      <p>CSV downloads will have the following format.</p>
-      <pre class="example-text">
+{% set form_title = "Download data for Observation 'GHI Instrument 1'" %}
```
alorenzo175: and here

sfa_dash/templates/forms/observation_upload_form.html
```diff
@@ -1,28 +1,25 @@
...
-	<div class="form-element">
-	  <label>My data is formatted in:</label><br>
-	  <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="csv" checked/>CSV
-      <input type="radio" data-toggle="collapse" data-target=".upload-example" name="file-format" value="json"/>JSON
-	</div>
-	<div class="form-element full-width">
-      <div class="upload-example csv show">
-		<p>Observation data in CSV format should follow the formatting of the example below.</p>
-	    <pre class="example-text">
+{% set form_title = "Upload data for Observation 'GHI Instrument 1'" %}
```
alorenzo175: and here

sfa_dash/templates/forms/cdf_forecast_download_form.html
alorenzo175: is it supposed to be set when templating? Looks like that isn't working at least when I run locally. Maybe this shouldn't be set at all in the form since you have the conditional block anyway

lboeman: You're right, but I don't know that it's necessary in the first place. I neglected to set this dynamically after moving away from the hard-coded values. If we need this subtitle to exist reminding a user that they're downloading the data for the site with it's metadata at the top of the page, all we need to do is print the name dynamically. Without it the page title is 'Download <data-type>' and the metadata is shown directly below that so it is redundant.

PR #46, opened 2019-04-16
Setup integration

lboeman:
Runs tests against the Development API, the AUTH0_CLIENT_SECRET is stored in Azure Pipelines in the variable group `dashboard-build-secrets`.


lboeman:
This may not work, Azure Pipelines doesn't want to supply the secret to the build even when 'Make secrets available to builds of forks' and 'Require a team member's comment before building a pull request' are set. Manually triggering a build within azure works though.

PR #48, opened 2019-04-17
add server side token storage

alorenzo175:
and adjust the error handler tests to not rely on auth0 at all which should avoid having to set CLIENT_ID and CLIENT_SECRET for now at least

### DIFF
--- a/sfa_dash/__init__.py
+++ b/sfa_dash/__init__.py
@@ -2,7 +2,7 @@ from flask import Flask, redirect, url_for, render_template, session, request
 from flask_seasurf import SeaSurf
 from flask_dance.consumer.storage.session import SessionStorage
 
-from sfa_dash.blueprints.auth0 import (make_auth0_blueprint, logout,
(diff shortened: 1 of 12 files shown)


sfa_dash/__init__.py
```diff
@@ -18,21 +19,28 @@ def create_app(config=None):
...
+        except ValueError:
+            return redirect(url_for('index'))
+        else:
+            if not authorized:
+                # means we have a token, not necessarily that it
+                # hasn't expired, but refreshing is handled
+                # by request_oauthlib and oauthlib
+                # and the api validates expiration
+                session['redirect_path'] = request.path
+                return redirect(url_for('auth0.login'))
```
lboeman: To retain the redirect after login, I think the except block needs to follow the same logic as the else here, otherwise it only redirects users who haven't been logged in recently.

alorenzo175:
yeah, we're trusting flask does the right thing with the cookies. I considered hashing the username w/ sha256 before storing, but I don't think it is private information that shouldn't go in a cookie

PR #50, opened 2019-04-17
Observation Time series plots

lboeman:
Adds general timeseries plots. Currently set up to display the last day of data by default but will load whatever data is available, so users can scroll backwards to see more data. This will have to be limitted to a reasonable amount of data at some point, but I'm holding off on setting that up because we don't yet have a method for retrieving the available range of data.

The current output of these plots is as below: 
Screenshot from 2019-04-17 16-28-25


wholmgren:
Nice!

lboeman:
I think this is good to go. I added plots to all of the currently applicable data types, perhaps there would be a plot for CDF forecast groups as well. I'm holding off on refactoring the copy pasta'd code to add bokeh_script and plot to the template arguments, because I think I could probably write a more general and useful helper once I have an idea of what the other types of plots will look like. 
I also added a notice to let users know when there is no data available, it's printed as below:
Screenshot from 2019-04-18 10-59-18

sfa_dash/plotting/timeseries.py
```diff
@@ -0,0 +1,119 @@
...
+    period_start = df.index[0]
+    period_end = df.index[-1]
+    # If there is more than a day of data, limit the default x_range
+    # to display only the most recent day of data. Users will be able
+    # to scroll to see past data.
+    interval_length = metadata['interval_length']
+    intervals_per_day = int(1440 / interval_length)
+    three_days = 3 * intervals_per_day
+    if df.index.size > three_days:
+        x_range_start = df.index[-three_days]
```
alorenzo175: df.index is a DatetimeIndex right? You should be able to use ``df.index[df.index.get_loc(period_end - pd.Timedelta('3d'), method='bfill')]``  and get rid of the interval length stuff and checkign the index size

alorenzo175: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.get_loc.html#pandas.Index.get_loc

sfa_dash/plotting/timeseries.py
```diff
@@ -0,0 +1,119 @@
...
+        key.
+
+    Returns
+    string
+        The appropriate x axis label.
+    """
+    variable = metadata['variable']
+    var_name = api_to_dash_varname(variable)
+    units = api_varname_to_units(variable)
+    label = f'{var_name}{units}'
```
alorenzo175: add a space between var_name and  units

sfa_dash/plotting/timeseries.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Returns
+    -------
+    string
+        The appropriate figure title.
+    """
+    site_name = metadata['site']['name']
+    start_string = start.strftime('%Y-%m-%d %H:%M')
+    end_string = end.strftime('%Y-%m-%d %H:%M')
+    figure_title = (f'{site_name} {build_x_axis_label(metadata)}'
```
alorenzo175: I think this should be the observation name and we probably don't need the variable/units there so ``f'{observation_name} from {start_string} to {end_string}'``

PR #54, opened 2019-04-22
Add a create button to obs, fx, cdf fx listings

lboeman:
Adds a button to the listing pages without site. "Create" buttons drop people on the Sites listing page with the query parameter `create=<type>` where type is 'observation', 'forecast' or 'cdf_forecast_group'
. Each Site's title will link to a creation form of the given type for that Site.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,5 +20,5 @@ setup(
         'cryptography',
         'blinker',
         'bokeh'
-        ]
+    ]
 )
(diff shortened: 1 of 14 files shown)


wholmgren:
Seems that we should have updated the text on the home page too:

Use the links on the left to access your data.

The Sites link will list the sites available to you. From this menu, you explore Sites and their associated data. Selecting a Site will display its metadata and present options to list or create Observations, Forecasts and Probabilistic Forecasts associated with the Site. ~This is currently the only menu from which you can create new Sites, Forecasts, Observations and Probabilistic Forecasts.~

The Observations, Forecasts and Probabilistic Forecasts links will list all of the items you have access to regardless of their Site association. ~However, you will not be able to Create new items from these pages.~

**add:** See documentation for more details.

PR #63, opened 2019-04-24
fix name of max_rotation_angle in site form and add it to keys to par…

### LINKED ISSUE #61: cannot make single axis tracker
Dashboard complained about the following error when I tried to make a single axis tracker: `max_rotation_angle: Value required when tracking_type='single_axis'`. I had specified 45 for that field.

lboeman:
…se for single axis tracking_type
closes #61


alorenzo175:
datamodel and api have since been changed to max_rotation_angle

PR #74, opened 2019-05-29
User admin

lboeman:
Basic User administration layout.
This PR adds a user administration seciton accessible from the 'Account' link in the upper right corner of the dashboard. It contains the following sections:

- User listing:  Displays users an administrator has access to. At the moment, this is currently just the Test User, who is also the only user in the organization. The /users/ endpoint only returns uuid, organization, created at and modified, so I used UUIDS in this version of the table.
user_listing

- User page: Displays user metadata and roles. Roles are listed by name and date added to the user.
user_view

- Role Listing: displays roles an administrator has access to. Lists name description and organization of the role. Note that an administrator might have access to view a role from another organization and view the permissions associated with it, but may not edit the role.
role_listing

- Role View: Display Role Metadata and permissions permissions are listed by their description and the date added to the Role.
role_view

- Permission Listing: Lists all permissions administrator has access to. Lists description, action, object_type and the applies_to_all for each permission. 
permission_listing

- Permission View: Displays metadata of the permission and the objects it allows access to. Perhaps if the permission has the `applies_to_all` flag set, objects should be hidden.
permission_view

These pages work with the users/permissions/roles endpoints added to the api, but the code is rudimentary and meant to showcase user interface more than provide a robust backend.

### DIFF
--- a/sfa_dash/__init__.py
+++ b/sfa_dash/__init__.py
@@ -83,8 +83,9 @@ def create_app(config=None):
 
     from sfa_dash.blueprints.main import data_dash_blp
     from sfa_dash.blueprints.form import forms_blp
+    from sfa_dash.blueprints.admin import admin_blp
 
-    for blp in (data_dash_blp, forms_blp):
+    for blp in (data_dash_blp, forms_blp, admin_blp):
         blp.before_request(protect_endpoint)
         app.register_blueprint(blp)
     return app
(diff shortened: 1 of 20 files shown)


lboeman:
Updated with a fixed table layout so column sizes don't jump when filtering. Wrapping shouldn't be an issue in fields other than 'Name' for now, might need adjusting down the road.

PR #85, opened 2019-06-03
correct use of first-child selector

lboeman:
The first-child selector I setup was too generic, and spilling over into the filter `<ul>` in the table header.


lboeman:
just going to roll this into the reports PR and merge since I have to fix conflicts and don't want to make Openshift build twice.

PR #93, opened 2019-06-04
Layout fixes

lboeman:
Should create more sane layouts on smaller screens.

- Stops date/time fields on download pages from overflowing out of the viewport by explicitly setting the width of form-group to 100%
- Stacks the sidebar on top of the rest of the content on small screens (767px wide and smaller).
- Also introduces an 'x' box in the organization filter, to close the dropdown.


lboeman:
Added the proposed timezone indication on `issue_time_of_day`. closes #95

PR #101, opened 2019-06-06
handle files with incorrect mimetype closes #100

### LINKED ISSUE #100: Handle unsupported mimetypes
https://sentry.io/organizations/solararbiter-rx/issues/1059259048/?project=1457208

This popped up in Justin's testing. The file he was uploading was labelled  'ArbiterTestFcst.csv' but had the mime-type 'application/vnd.ms-excel' and the checking is faulty.

lboeman:
Reloads the upload form with an error about unsupported file type when a user tries to upload anythong other than text/csv or application/json. This may need to be investigated further as it looks like the original error was from trying to upload a .csv file with a different mimetype. Perhaps this was due to being created with excel and we should support that?

### DIFF
--- a/sfa_dash/blueprints/form.py
+++ b/sfa_dash/blueprints/form.py
@@ -296,7 +296,10 @@ class UploadForm(BaseView):
         posted_file = request.files[f'{self.data_type}-data']
         if posted_file.filename == '':
             return 'no filename'
-        if posted_file.mimetype == 'text/csv':
+        if (
+                posted_file.mimetype == 'text/csv' or
+                posted_file.mimetype == 'application/vnd.ms-excel'
+        ):
             posted_data = posted_file.read()
             post_request = self.api_handle.post_values(
                 uuid,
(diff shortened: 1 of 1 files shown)


alorenzo175:
Looks like the file is a valid csv but not sure how the encoding is decided. Is it by the browser? 

In any case, might be better to just assume anything that isn't json is a csv. Then the api can complain if has an issue

wholmgren:
I suggest we explicitly attempt to parse `application/vnd.ms-excel` as csv and if it fails then we report that it must be a csv file.

wholmgren:
Some googling suggests that this is a common issue for windows users that have excel. Windows is bad.

alorenzo175:
> I suggest we explicitly attempt to parse application/vnd.ms-excel as csv and if it fails then we report that it must be a csv file.

I think the logic that says it must be a csv has to go in the API because it does the actual parsing of the file and right now we probably can't tell errors due to bad csv fields from errors because the file is actually an excel file.

Issue #102, opened 2019-06-06
inconsistent search bar behavior when using browser back button


wholmgren:
From @MrWindAndSolar... When clicking the back button from the obs section of the dashboard, the search persists in the box but is inconsistent with the display.

lboeman:
closed by #107

PR #107, opened 2019-06-06
Search persistence

lboeman:
Remedies #102 but revealed I need to fix the filter/search functionality to apply all filters on change. Currently, the filters apply individually without respect to eachother.

For example, deselecting `Reference` from the organization filter hides all that data, and then searching for 'GHI' hides everything that does not contain 'GHI'. Removing the search sets visible=true on everything, without checking the org filter.


lboeman:
Rewrote this such that on every filter change, it reapplies all filters, so that we don't enter a broken state when one filter being applied/removed affects elements affected by other filters.

sfa_dash/static/js/table-search.js
```diff
@@ -1,29 +1,107 @@
...
-
-  $(".results tbody tr:containsi('" + searchSplit + "')").each(function(e){
-    $(this).attr('visible','true');
-  });
-
-  var jobCount = $('.results tbody tr[visible="true"]').length;
-    $('.counter').text(jobCount + ' item');
+    /* Add containsi filter to JQuery selectors that selects for any word in a phrase */
+    $.extend($.expr[':'], {'containsi': function(elem, i, match, array){
+            return (elem.textContent || elem.innerText || '').toLowerCase().indexOf((match[3] || "").toLowerCase()) >= 0;
```
wholmgren: I don't fully understand this statement. might help to assign `(elem.textContent || elem.innerText || '').toLowerCase()` to its own variable. then `newvar.indexOf((match[3] || "").toLowerCase()` to another variable. hoping the variable names make it self documenting. also what is `match[3]`?

sfa_dash/static/js/table-search.js
```diff
@@ -1,29 +1,107 @@
...
+        var jobCount = $('.results tbody tr[visible="true"]').length;
+        if(jobCount == '0') {$('.no-result').show();}
+        else {$('.no-result').hide();}
+    }
+    toHideFns.push(searchTable);
+    /*
+     * If the Provider header id exists, build a dropdown ul of checkboxes from the
+     * list of organizations in the table.
+     */
+    if ($('#provider-header').length){
```
wholmgren: looks like this is mostly moved from the other file, so ok to leave as is for now. but we'll probably want to build checkbox drop down lists for at least variable, maybe others.

sfa_dash/static/js/table-search.js
```diff
@@ -1,29 +1,107 @@
...
+        var availableOrgs = new Set([]);
+        var orgs = $(".provider-column")
+        for (i = 0; i < orgs.length; i++) {
+            availableOrgs.add(orgs[i].textContent);
+        }
```
wholmgren: I think either jquery or d3 has a function that will compute this set for you.

PR #108, opened 2019-06-24
Reports api

lboeman:
Integrate api into reports pages.


lboeman:
Started to incorporate core's datamodel and api. Will still need the original api interface for the time being, because the Core does not have delete methods or cdf_forecast methods yet.

alorenzo175:
The function to call to generate the report HTML is https://github.com/SolarArbiter/solarforecastarbiter-core/blob/41c9de2e966662c047ddf431f6d556ee47e152bf/solarforecastarbiter/reports/main.py#L247

sfa_dash/blueprints/reports.py
```diff
@@ -132,13 +136,13 @@ class ReportView(BaseView):
...
-        if report_request.status_code == 200:
-            self.metadata = report_request.json()
-        else:
-            abort(404)
+        token = oauth_request_session.token['access_token']
+        session = APISession(token)
+        try:
+            report_request = session.get_report(uuid)
+        except HTTPError as e:
+            abort(404);
```
lboeman: @alorenzo175  This isn't complete or anything, but was this sort of inlining what you were suggesting? Or did you mean keeping it in the api_interface module?

lboeman:
Updated the start/end inputs to be a regular text box with a placeholder of 'YYYY-MM-DDTHH:MMZ', and added a javascript validator to warn people of improperly formatted datetimes on submission. 
Screenshot:
Screenshot from 2019-06-28 09-55-02

Does not check for range, e.g. you can enter a month of 13 and it will post to the API but you will still get an invalid datetime error back.

The date time widget is also used on the download form so if we go this route, perhaps that should be updated as well.

wholmgren:
> Updated the start/end inputs to be a regular text box with a placeholder of 'YYYY-MM-DDTHH:MMZ', and added a javascript validator to warn people of improperly formatted datetimes on submission.

Sounds good for now. We're going to need to clean up time/timezone specs in a handful of places later on, so don't over do it now.

PR #114, opened 2019-07-18
Reports widget update

lboeman:
Makes the Report observation and forecast pair widget less frustrating to use. Has a persistent list of observations. When an observation selection is made, the forecasts list is limited by site and variable.


lboeman:
@wholmgren made those minor changes and tested against the prod db, the page loads slowly due to the requests for obs & fx, but the select inputs are responsive.

wholmgren:
the slow page loading is due to the network/server delays rather than initial rendering? If so, go ahead and merge.

Issue #117, opened 2019-08-06
Data sharing interface


lboeman:
The initial design of the data sharing interface should parallel the manual controls exposed by the API. Simpler call-to-action type interface should be implemented later.

lboeman:
Running into need for a user information type of endpoint for the API, to filter RBAC objects by the current admin's organization such that they don't run into errors trying to add other organization's objects to their RBAC objects.

wholmgren:
what's the status of this? Seems like it might be complete.

lboeman:
I think this can be closed and specific issues can be opened moving forward. I think the basic functionality is complete here.

PR #118, opened 2019-08-09
User management

lboeman:
Updates and exposes RBAC dashboard features. Working towards addressing #117 
Still in rough shape, and will depend on an API update to function correctly.

### DIFF
--- a/sfa_dash/api_interface/permissions.py
+++ b/sfa_dash/api_interface/permissions.py
@@ -1,4 +1,4 @@
-from sfa_dash.api_interface import get_request
+from sfa_dash.api_interface import get_request, post_request, delete_request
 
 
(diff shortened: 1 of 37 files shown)


lboeman:
Updated this to include some basic UI for the features introduced by https://github.com/SolarArbiter/solarforecastarbiter-api/pull/145. 

- Role page has permissions and users tabs.
role_users
role_permissions
- Form for granting role to a single user by id. To be upgraded to allow using email instead after implementing https://github.com/SolarArbiter/solarforecastarbiter-api/pull/145.
role_grant

- Update form for removing roles from users to redirect to the role view when necessary, and also display a warning when the administrator does not have permissions to view the user.
unreadable_user_error

PR #122, opened 2019-09-27
add start,end query params to fx/obs/cdf forecast pages

### LINKED ISSUE #120: Restrict start when retrieving observation values for plots
Since there is no reason to get all the data if only the last 3 days are shown

lboeman:
closes #120 
adds start, end query params to the forecast pages and defaults to start=-3days, end=today.

### DIFF
--- a/sfa_dash/blueprints/dash.py
+++ b/sfa_dash/blueprints/dash.py
@@ -4,6 +4,7 @@ on site-section.
 from sfa_dash.blueprints.base import BaseView
 from sfa_dash.api_interface import sites
 from flask import render_template, request, url_for
+import pandas as pd
 
 
 class DataDashView(BaseView):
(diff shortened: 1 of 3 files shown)


sfa_dash/blueprints/dash.py
```diff
@@ -20,6 +21,29 @@ def get_site_metadata(self, site_id):
...
+        start,end
+            Tuple of ISO 8601 datetime strings representing the start, end.
+        """
+        # set default arg to an invalid timestamp, to trigger ValueError
+        start_arg = request.args.get('start', 'x')
+        end_arg = request.args.get('end', 'x')
+        try:
+            start = pd.Timestamp(start_arg)
+        except ValueError:
+            start = pd.Timestamp.now() - pd.Timedelta('3days')
```
alorenzo175: should use ``.utcnow()`` and avoid ambiguities with leaving off the timezone

lboeman:
Yeah, this should be working. I'll add open an issue to add ui for the start/end for the query parameters in the future.

PR #138, opened 2019-10-08
Site perm 500

lboeman:
Handles partial permissions, leaning towards restricting data access. I don't think this is a good long term solution, but a stop gap to some of the errors we've seen recently. Hoping to discuss how we want to handle the possibilities of handling the possibility of impractical state of permissions in the framework. 

This PR does the following:
- Allows listing pages (e.g. `/observations`) to display all readable observations. In the case where the associated site is unreadable, displays 'Site Unavailable'. 
- Catches the error thrown when a user navigates to a listing for a site they do not have access to (e.g. https://dashboard.solarforecastarbiter.org/observations/?uuid=933bc5b8-7e49-11e9-b650-0a580a8003e9 with no access to the UO SRML Ashland, OR site).
 
This does have the side effect of responding with a 404 message when accessing valid data where the associated site or values is not readable. 

Moving forward, I think the dashboard should separately handle data requests and respond to each request with as much information as is available to the user. For instance, viewing an observation for which you lack `read_values` and `read` on the associated site should still yield metadata, and perhaps the uuid of the site (available in the observation metadata) and a message about lack of read_value permissions. This does put a damper on my plans to use the core api module going forward, as it does the work of inserting site metadata into the data objects and raises an HTTPError when either read permission is missing.

I was thinking of creating a spreadsheet mapping permission actions to object types and what practical effects they have in the framework. Might be helpful in creating better documentation for the RBAC system.

related to #135 #136

### DIFF
--- a/sfa_dash/blueprints/util.py
+++ b/sfa_dash/blueprints/util.py
@@ -312,6 +312,7 @@ def handle_response(request_object):
         handled.
     """
     if not request_object.ok:
+        errors = {}
         if request_object.status_code == 400:
             errors = request_object.json()
         elif request_object.status_code == 401:
--- a/sfa_dash/blueprints/util.py
+++ b/sfa_dash/blueprints/util.py
@@ -295,14 +295,21 @@ def handle_response(request_object):
 
     Returns
     -------
-    dict
-        The parsed json of a response.
+    dict, str or None
+        Note that this function checks the content-type of a response
+        and returns the appropriate type. A Dictionary parsed from a
+        JSON object, or a string. Returns None when a 204 is encountered.
+        Users should be mindful of the expected response body from the
+        API.
 
     Raises
     ------
     sfa_dash.errors.DataRequestException
         If a recoverable 400 level error has been encountered.
         The errors attribute will contain a dict of errors.
+    requests.exceptions.HTTPError
+        If the status code received from the API could not be
+        handled.
     """
     if not request_object.ok:
         if request_object.status_code == 400:
@@ -333,10 +340,13 @@ def handle_response(request_object):
             # so that they can be reported to Sentry.
             request_object.raise_for_status()
     if request_object.request.method == 'GET':
+        # all GET endpoints should return a JSON object
         if request_object.headers['Content-Type'] == 'application/json':
             return request_object.json()
         else:
             return request_object.text
+    # POST responses should contain a single string uuid of a newly created
+    # object unless a 204 No Content was returned.
     if request_object.request.method == 'POST':
         if request_object.status_code != 204:
             return request_object.text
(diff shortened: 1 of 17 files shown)


lboeman:
Corrected such that when a site is unavailable, the listing pages display 'Site Unavailable' in the site column of a given table. Like so: 
Screenshot from 2019-10-22 13-40-30

And on a obs/fx/cdf page, the site id is displayed with a message. Plots are not displayed because core currently uses sites for plotting.
Screenshot from 2019-10-22 13-40-38

I've also updated this with some better documentation in `blueprints/base.py` and `blueprints/main.py`. Did a small amount of refactoring in main and created a method in BaseView, `get_breadcrumb_dict` with documentation on how it should be overridden for creating breadcrumbs. It has a counterpart method `breadcrumb_html` that accepts it's return value and creates the html for templates. The idea being that BaseView should supply the formatting logic, and a well-documented method for override to supply input to the formatter. I think this pattern can be implemented to reduce most of the copy pasta in the dashboard, and simplify working on the dashboard.

sfa_dash/blueprints/util.py
```diff
@@ -299,14 +331,22 @@ def handle_response(request_object):
                     f' <a href="{previous_page}">Return to the previous '
                     'page.</a>')
             raise DataRequestException(request_object.status_code, **errors)
-        # Other errors should be due to bugs and not by attempts to reach
-        # inaccessible data. Allow exceptions to be raised
-        # so that they can be reported to Sentry.
+        elif request_object.status_code == 422:
+            errors = {'422': ['Failed to compute aggregate Values']}
+            raise DataRequestException(request_object.status_code, **errors)
```
alorenzo175: could remove this copy pasta by setting ``errors = {}`` above, removing the else and then have 
```python
if errors:
    raise DataRequestException(request_boject.status_code, **errors)
else:
    request_object.raise_for_status()
```

sfa_dash/templates/data/asset.html
```diff
@@ -10,13 +10,14 @@
 {% if delete_link is defined %}
 <a role="button" class="btn btn-danger btn-sm" href="{{ delete_link }}">Delete</a>
 {% endif %}
-{% include "sections/notifications.html" %}
 <script src="/static/js/plot-bound-parsing.js"></script>
```
alorenzo175: is this needed if no plot?

sfa_dash/templates/data/asset.html
```diff
@@ -10,13 +10,14 @@
...
-{% if plot is defined %}
+{% include "sections/notifications.html" %}
+{% if plot is not none %}
 <div class="row data-plots-wrapper">
 {{ plot | safe }}
 </div>
 {% endif %}
+{% if plot is not none %}
```
alorenzo175: why not just remove the endif one line above?

PR #146, opened 2019-10-28
Dash tests

lboeman:
Adds some tests to check that all the available routes return 200 to get requests. I need to figure out establishing a test database and API instance in azure for these tests to work. Then it would be simple to add selenium testing and what not.


sfa_dash/tests/views/test_success.py
```diff
@@ -0,0 +1,53 @@
+from sfa_dash.conftest import BASE_URL
+
+
+def test_get_no_arg_routes(client, no_arg_route):
+    resp = client.get(no_arg_route, base_url=BASE_URL)
+    assert resp.status_code == 200
+
+
+def test_site_filtered_no_args(client, no_arg_route, site_id):
```
alorenzo175: I'm confused what this does for no_arg_route

lboeman: The the site_id is used for accessing the obs/fx/cdf at a site. I guess no_arg_route is kind of a misnomer, but there are no path parameters.

PR #147, opened 2019-10-29
start dashboard aggregate pages

lboeman:
Adding aggregate functionality to the dashboard.


lboeman:
Setup the aggregate listing page, individual pages, creation form and observation addition form. 
@wholmgren  @alorenzo175 I'm trying to think of how to handle setting `effective_until` on observations. Currently, the observation addition form is set up as such: 
obs_addition
A user just selects the observations they'd like to add and enters an `effective_from` date. How do you feel about reusing  this pattern for setting `effective_until`?

I am trying to think of a better way to decide exactly which observations to display on which page, currently I'm displaying everything with the same variable and shorter interval_length.

wholmgren:
Effective from/until are most likely unique to each observation, so I think it will be confusing to make them apply to all selected like this. Maybe put those in their own columns? If not set, then we assume effective from first through last data point.

wholmgren:
Maybe we should state priorities. I suggest:
1. We should get it done.
2. We should make it easy to add lots of observations to an aggregate.
3. We should make it possible to set unique effective time ranges for each observation in an aggregate.

alorenzo175:
``effective_until`` should only apply to observations that are already part of the aggregate. So, the aggregate page should a list of associated observations and there have the option to set ``effective_until`` on each individually.

I think the batch like you have for ``effective_from`` is OK, but you should probably exclude any observations from the list that are already in the aggregate and have no ``effective_until``set.

lboeman:
I just realized I've completely missed implementing aggregate forecasts here. @wholmgren, @alorenzo175 Are there any requirements/restrictions of the forecasts made for an aggregate?

lboeman:
This grew into a pretty big PR trying to handle aggregates across the dashboard. I've tried to summarize updates below.

1. Reduced sidebar navigation to sites, aggregates, reports, observations and forecasts. After some discussion with @wholmgren this seemed reasonable and easier to document. This removes the need for the intermediate 'select a site' page when creating new obs/cdf/fx. List pages remain the same when accessed via a site, but now the pages that display all of the users available data are sequestered in their own section with only tab navigation like so: 
data_listing
Note that for forecasts, the last column is now 'Site/Aggregate' and observations remain as 'Site'
2. Forecasts (including probabilistic) that reference aggregates inherit the variable of the aggregate they reference. Forms are as follows:
agg_fx_form
3. Aggregate observation tables display a link to set the `effective_until` field for each observation on the aggregate if not already present.
aggregate_page
4. Added form for setting `effective_until` on a single aggregate. 
effective_until
5. Handle orphaned objects (which can only be forecasts at the moment) by displaying 'Object Deleted' in the site field when both `aggregate_id` and `site_id` are none.
6. Haven't yet implemented Aggregate forecast listing, as this isn't available from the API at the moment. This could be implemented on the dashboard if need be, but it seems like a feature we would want on the api. Api issue: https://github.com/SolarArbiter/solarforecastarbiter-api/issues/179

wholmgren:
looks good to me. Is there anything left to review?

lboeman:
I don't think so. It's probably more efficient to do any further review on a running dashboard and adjust from there.

sfa_dash/blueprints/main.py
```diff
@@ -151,31 +288,27 @@ def get(self, uuid, **kwargs):
...
-                        'Site Access': [
-                            'Site inaccessible. Plots will not be displayed.']
-                    },
-                    'plot': None,
-                })
+                self.temp_args.update({'plot': None})
             else:
                 # If site data is available, try to create a plot
-                self.insert_plot(uuid, start, end)
+                if self.metadata.get('site') is not None:
```
alorenzo175: should not have to check this once ``blueprints/util.py`` can handle either site or agg

sfa_dash/blueprints/main.py
```diff
@@ -69,6 +197,15 @@ def get_breadcrumb_dict(self):
             breadcrumb_dict[f'{self.human_label}s'] = url_for(
                 f'data_dashboard.{listing_view}',
                 uuid=self.metadata['site_id'])
+        elif self.metadata.get('aggregate') is not None:
```
alorenzo175: I don't think aggregates logically belong with forecasts. They are more like observations without sites.

sfa_dash/blueprints/dash.py
```diff
@@ -21,6 +22,75 @@ def template_args(self, **kwargs):
...
+
+        Returns
+        -------
+        link_html: str
+            An anchor tag referencing the site/aggregates page on the
+            dashboard or a string uuid if the metadata wasn't not available,
+            indicating that the current user does not have access to the
+            site or aggregate.
+        """
+        if self.metadata['site_id'] is not None:
```
alorenzo175: if you use ``self.metadata.get('site_id')`` here and below for aggregate_id, this should work just as well for observations

sfa_dash/blueprints/dash.py
```diff
@@ -21,6 +22,75 @@ def template_args(self, **kwargs):
...
         return handle_response(sites.get_metadata(site_id))
 
+    def set_site_or_aggregate_metadata(self):
+        """Searches for a site_id or aggregate_id  in self.metadata
+        and loads the expected metadata object from the api in either
+        the 'site' or 'aggregate' key. If the object could not be retrieved,
+        sets a warning, sets the 'plot' template argument to None and
+        reraises the DataRequestError.
+        """
+        if self.metadata['site_id'] is not None:
```
alorenzo175: see comment below about using ``get`` so that this also works for observations

sfa_dash/blueprints/main.py
```diff
@@ -23,6 +25,137 @@ class SingleObjectView(DataDashView):
     """
     template = 'data/asset.html'
 
+    def __init__(self, data_type):
```
alorenzo175: is it really necessary to separate this out. Seems like a little careful handling of getting site/aggregate_id would avoid the extra copy pasta

sfa_dash/blueprints/util.py
```diff
@@ -25,35 +25,69 @@ def creation_link(cls, data_type, site_id=None):
...
-            if id_key == 'forecast_id':
-                table_row['link'] = url_for('data_dashboard.forecast_view',
-                                            uuid=data[id_key])
+            table_row['link'] = url_for('data_dashboard.observation_view',
+                                        uuid=data['observation_id'])
+            table_rows.append(table_row)
+        return table_rows
+
+    @classmethod
+    def create_table_elements(cls, data_list, view_name, **kwargs):
```
alorenzo175: seems like these don't need to be separated. 

before the loop, if observations, don't get aggregates, and set the no location text ('Site unavailable').
Then modify the 
```python
if data.get('site_id') is not None:
   blah
elif data.get('aggregate_id') is not None:
   blah
else:
   location_id = None
```

test_api/Dockerfile
```diff
@@ -3,3 +3,7 @@ RUN git clone https://github.com/SolarArbiter/solarforecastarbiter-api.git
 WORKDIR solarforecastarbiter-api
 RUN pip install -r requirements.txt
 RUN pip install .
+RUN git clone https://github.com/SolarArbiter/solarforecastarbiter-core.git
```
alorenzo175: I don't think this is necessary now that the setup.py has been fixed

sfa_dash/blueprints/main.py
lboeman: You're right. I think before I arrived at some of the refactoring I did this seemed reasonable but this could probably all be combined again.

sfa_dash/blueprints/main.py
lboeman: I think having the breadcrumb relating the forecast to the Aggregate is helpful. That relationship is duplicated in the metadata section as well though, so I'm not completely attached to it.

alorenzo175: whoops. I didn't look carefully and thought this was for data_type == 'aggregate' in the __init__. I don't have a problem with keeping this breadcrumb

lboeman:
Alright, I've removed some of that copy pasta and setup plotting across all the data types. The Aggregate pages are looking like this now:
Screenshot from 2019-11-07 16-00-28

PR #150, opened 2019-11-12
Add email to user views and tables

lboeman:
Adds emails to tables when available. Changes the "Grant role" page to use email rather than user id.
If we want user email/org to appear on the role's user listing, we'll have to insert it at the API or expose an endpoint with all the user data as discussed in https://github.com/SolarArbiter/solarforecastarbiter-api/issues/154

### DIFF
--- a/sfa_dash/blueprints/admin.py
+++ b/sfa_dash/blueprints/admin.py
@@ -216,6 +216,14 @@ class RoleListing(AdminView):
 class RoleView(AdminView):
     template = 'forms/admin/role.html'
 
+    def get_email(self, user_id):
+        try:
+            email = handle_response(users.get_email(user_id))
+        except DataRequestException:
+            return 'Email Unavailable'
+        else:
+            return email
+
     def get(self, uuid, **kwargs):
         role_table = request.args.get('table', 'permissions')
         try:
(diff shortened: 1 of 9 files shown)


lboeman:
Force pushed to remove unnecessary commits from trying to debug. @alorenzo discovered the issue in failing api calls when auth0 client id and secret weren't set for the api in the testing containers. The failures are handled in the api now, rather than requiring providing secrets for testing.

sfa_dash/blueprints/admin.py
```diff
@@ -233,11 +241,14 @@ def get(self, uuid, **kwargs):
...
-                                     k, {}).get('organization', '')}
-                             for k, v in role['users'].items()}
+            role['users'] = {
+                uid: {'user_id': uid,
+                      'added_to_user': user,
+                      'email': user_map.get(uid, {}).get(
+                          'email', self.get_email(uid)),
+                      'organization': user_map.get(uid, {}).get(
+                          'organization', 'Organization Unavailable')}
+                for uid, user in role['users'].items()}
```
alorenzo175: As I was recently reminded, even though we might not use the result of ``self.get_email(uid)``  the function is still executed and makes an unnecessary call to the API for every uid. We probably want to write out the for loop to get the email from user_map first, and if it isn't found call ``self.get_email``.

```python
for uid, user in role['users'].items():
    email = user_map.get(uid, {}).get('email', None)
    if email is None:
        email = self.get_email(uid)
    ...
```

PR #151, opened 2019-11-13
Aggregate forecast listings

lboeman:
Adds ability to list forecasts or cdf forecasts made for an aggregate, and adds the same subnav as found on sites to aggregates except that the 'observations' link returns the aggregate page, where the listing of constituent observations can be found. 
Like so:
Screenshot from 2019-11-13 11-06-16

### DIFF
--- a/sfa_dash/blueprints/aggregates.py
+++ b/sfa_dash/blueprints/aggregates.py
@@ -6,7 +6,6 @@ import pandas as pd
 
 from sfa_dash.api_interface import observations, sites, aggregates
 from sfa_dash.blueprints.base import BaseView
-from sfa_dash.blueprints.data_listing import DataListingView
 from sfa_dash.blueprints.util import (filter_form_fields, handle_response,
                                       parse_timedelta, flatten_dict)
 from sfa_dash.errors import DataRequestException
@@ -276,6 +275,11 @@ class AggregateView(BaseView):
     api_handle = aggregates
     plot_type = 'aggregate'
     human_label = 'Aggregate'
+    subnav_format = subnav_format = {
+        '{observations_url}': 'Observations',
+        '{forecasts_url}': 'Forecasts',
+        '{cdf_forecasts_url}': 'Probabilistic Forecasts',
+    }
 
     def get_breadcrumb_dict(self):
         breadcrumb_dict = OrderedDict()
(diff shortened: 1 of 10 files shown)


sfa_dash/blueprints/aggregates.py
```diff
@@ -292,6 +310,7 @@ def set_template_args(self):
             'breadcrumb': self.breadcrumb_html(
                 self.get_breadcrumb_dict()),
             'aggregate': self.metadata,
+            'subnav': self.format_subnav(**self.get_subnav_kwargs())
```
alorenzo175: Passing dicts around like this makes it really hard to track down exactly what is happening. Why not pass the keywords here and get rid of ``get_subnav_kwargs`` completely?

sfa_dash/blueprints/util.py
```diff
@@ -150,9 +153,20 @@ def get_forecast_table(cls, site_id=None, **kwargs):
...
+        elif aggregate_id is not None:
+            creation_link = cls.creation_link('forecast',
+                                              aggregate_id=aggregate_id)
+            request_kwargs['aggregate_id'] = aggregate_id
+        else:
+            creation_link = None
+
         forecast_data = handle_response(
-            forecasts.list_metadata(site_id=site_id))
+            forecasts.list_metadata(**request_kwargs))
```
alorenzo175: passing the dict like this obfuscates what's going on. You really don't need request_kwargs at all I think and can just do ``forecasts.list_metadata(site_id=site_id, aggregate_id=aggregate_id)

sfa_dash/blueprints/util.py
```diff
@@ -185,9 +199,20 @@ def get_cdf_forecast_table(cls, site_id=None, **kwargs):
...
+        elif aggregate_id is not None:
+            creation_link = cls.creation_link('cdf_forecast_group',
+                                              aggregate_id=aggregate_id)
+            request_kwargs['aggregate_id'] = aggregate_id
+        else:
+            creation_link = None
+
         cdf_forecast_data = handle_response(
-            cdf_forecast_groups.list_metadata(site_id=site_id))
+            cdf_forecast_groups.list_metadata(**request_kwargs))
```
alorenzo175: see above, don't need request_kwargs

sfa_dash/blueprints/data_listing.py
```diff
@@ -33,60 +36,87 @@ def __init__(self, data_type, **kwargs):
...
-        return breadcrumb
+            subnav_kwargs['observations_url'] = url_for(
+                'data_dashboard.observations', **kwargs)
+        subnav_kwargs['forecasts_url'] = url_for(
+            'data_dashboard.forecasts', **kwargs)
+        subnav_kwargs['cdf_forecasts_url'] = url_for(
+            'data_dashboard.cdf_forecast_groups', **kwargs)
+        return subnav_kwargs
 
     def get_template_args(self, **kwargs):
```
alorenzo175: so kwargs is only ever aggregate_id or site_id right? I think it's more clear if arguments are explicit ``get_template_args(self, site_id=None, aggregate_id=None)`` so the rest of the function can use these. As far as I can tell this class isn't subclassed anywhere where this might be a problem, right?

PR #166, opened 2019-11-20
allow forecast-aggregate pair selection on report form

lboeman:
Adds a radio button that allows switching from picking observation-forecast pairs to aggregate-forecast pairs. Currently updated to parse metadata into the format proposed in https://github.com/SolarArbiter/solarforecastarbiter-core/pull/256: 
`[{'forecast': <forecast_id>,
   'observation': <observation_id>}]`

Screenshots of both states of the new interface below.
Screenshot from 2019-11-20 10-50-22
Screenshot from 2019-11-20 10-50-30
This PR should wait for the core PR and ensure it works correctly before being merged.

### DIFF
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -1,3 +1,4 @@
+import pdb
 """Draft of reports endpoints/pages. Need to integrate core report generation.
 """
 from flask import request, redirect, url_for, render_template
@@ -65,8 +66,9 @@ class ReportForm(BaseView):
         # appropriately indexed forecast.
         truth_ids = filter_form_fields('truth-id-', form_data)
         truth_types = filter_form_fields('truth-type-', form_data)
-        pair_list = list(zip(fx, truth_types, truth_ids))
-        pairs = [{'forecast': p[0], p[1]: p[2]} for p in pair_list]
+        pdb.set_trace()
+        pairs = [{'forecast': f, truth_types[i]: truth_ids[i]}
+                 for i, f in enumerate(fx)]
         return pairs
 
     def parse_metrics(self, form_data):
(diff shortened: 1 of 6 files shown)


lboeman:
The forecast-> aggregate pairing ended up being simple enough since the only valid pairs are between a forecast and the aggregate they reference. 

However, I did discover that report computation fails when the forecast interval_length is shorter than the observation/aggregate. This happens to be the case with the test aggregates having interval_lengths of 1 hour and the test aggregate forecasts at 5 minutes. 

I've added validation here at the dashboard "aggregate forecast" creation form, to limit the interval_length to be greater than the aggregates. Is this something that's reasonable, or should I allow any forecast interval_length and limit the selection of aggregates at report creation?

lboeman:
Finally came to the reasonable conclusion that Observations/Aggregates should just be filtered on the reports creation page based on the currently selected forecast's interval length. This enforces the policy that for a report object_pair to be valid, the forecast must have a longer interval length than the observation.
This adds the interval-length as a data-* attribute to all of the obs/fx/agg <option> elements, and js for filtering.

sfa_dash/blueprints/reports.py
```diff
@@ -49,13 +54,19 @@ def template_args(self):
...
-        obs = filter_form_fields('observation-id-', form_data)
-        pairs = list(zip(fx, obs))
+
+        # observations and aggregates are passed in as truth-id-{index}
+        # and truth-type-{index}, so we must match these with the
+        # appropriately indexed forecast.
+        truth_ids = filter_form_fields('truth-id-', form_data)
+        truth_types = filter_form_fields('truth-type-', form_data)
+        pair_list = list(zip(fx, truth_types, truth_ids))
+        pairs = [{'forecast': p[0], p[1]: p[2]} for p in pair_list]
```
alorenzo175: probably more clear to just loop through enumerate forecasts

```python
pairs = [{'forecasts': f, truth_types[i]: truth_ids[i]} for i, f in enumerate(fx)]
```

PR #169, opened 2019-11-25
Add metrics/categories to report forms

lboeman:
Depends on https://github.com/SolarArbiter/solarforecastarbiter-core/pull/224

This PR will introduce metrics and categories( formerly referred to as subdivision in draft versions?) into reports selection. Currently in draft state and drawing from the metrics modules to populate options. 

Very basic widgets look as follows: 
Screenshot from 2019-11-25 15-37-21

@wholmgren Is the default behavior to compute all permutations of the selected metric and category for each object pair?

### DIFF
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -69,13 +69,6 @@ class ReportForm(BaseView):
                  for i, f in enumerate(fx)]
         return pairs
 
-    def parse_metrics(self, form_data):
-        """Collect the keys (name attributes) of the form elements with a value
-        attribute of metrics. These elements are checkbox inputs, and are only
-        included in the form data when selected.
-        """
-        return [k.lower() for k, v in form_data.items() if v == 'metrics']
-
     def parse_filters(self, form_data):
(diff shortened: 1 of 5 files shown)


wholmgren:
Yes, that was my idea, though I'm open to alternatives if you can make a good interface without too much effort.

lboeman:
> Yes, that was my idea, though I'm open to alternatives if you can make a good interface without too much effort.

I think that's a good place to start. I just wanted to check that I had the right behavior in mind. I think with the report sorting interface we discussed this will provide most of the desired output with a few clicks on the rendered report.

lboeman:
@wholmgren Is there a combination of metrics/categories that should be checked by default? I realized that some metrics require a reference forecast, so perhaps those are not included by default.

wholmgren:
> Is there a combination of metrics/categories that should be checked by default? 

For now, maybe let's check MAE, MBE, RMSE and all of the categories.

lboeman:
Updated version with the human readable labels:
Screenshot from 2019-12-09 12-53-54

PR #174, opened 2019-12-07
Add signature and html download path

alorenzo175:
Added the download path (/reports/<uuid>/html) but no link in report yet. Downloads a zip file with the GPG signature, report, and various checksums.


alorenzo175:
closes https://github.com/SolarArbiter/solarforecastarbiter-core/issues/157 and makes progress on https://github.com/SolarArbiter/solarforecastarbiter-core/issues/158

alorenzo175:
For now, the signing happens on the dashboard. I do think we might consider moving report view (html body, full html, pdf, etc) to the API and have a background job handle it and avoid blocking any flask threads.

alorenzo175:
Ah I meant moving the final report templating and signing to the API since it may take some time to do things like create the final PDF of a report. Hashing/signing is quick.

lboeman:
Ah okay :+1:

PR #175, opened 2019-12-10
Data sharing issues

### LINKED ISSUE #161: aggregate 422
two questions:

1. why is there a 422?
2. can we display more helpful information to the user?

Screen Shot 2019-11-18 at 11 58 52 AM

### LINKED ISSUE #173: minor issues with data sharing interface
* "Create a new permission" should be "Create a New Permission" for consistency with "Existing Permissions". 
* Seems like Sites should be first in the list.
* "Check to allow the action on all site in your organization." should say all sites
* the applies to all option should indicate that it will automatically update when new sites, for example, are added.
* is update actually allowed for sites, observations, and forecasts?
* min table height should be smaller when there are not many sites visible. the submit button was hiding from me off the bottom of the page and I briefly wondered how to proceed.
* on my mac firefox it's not obvious that the table is scrollable until you try to scroll. maybe nothing we can do about that.
* would be nice if the roles table allowed you to select/deselect action and object types (like we do for organization on the forecast & obs page)
* super annoying that I went through the tedious process of selecting my permissions only to find the page has been wiped because I dared to put special characters in the role name `read will's obs, read/write fx`
* I wonder if it would be easier to force people to create a role, then add permissions to it
* Weird that I could grant role to `justin@sharply-focused.com` but his organization is listed as "Organization Unavailable". I get that there are probably good reasons for it, but it's just weird that it shows up that way.
* more vertical white space between headings, please.

lboeman:
closes #173 
closes #161 
Handles a list of small issues all found on the "User Administration" pages. The biggest change in this PR is that users must create a role before adding permissions. This side steps some issues with persisting selected permissions on form failure.

### DIFF
--- a/sfa_dash/templates/forms/admin/permissions_form.html
+++ b/sfa_dash/templates/forms/admin/permissions_form.html
@@ -13,6 +13,9 @@
 {% elif data_type == 'report' %}
   {% set row_macro = table_macros.report_row %}
 {% else %}
+{% if form_data is not defined %}
+  {% set form_data = {} %}
+{% endif %}
 {% set row_macro = table_macros.data_row %}
 {% endif %}
 {% block fields %}
@@ -37,7 +40,7 @@
 <div class="form-element full-width">
   <label>Applies to all</label>
   <input type="checkbox" name="applies-to-all" data-toggle="collapse" data-target="#object-selector">
-  <p>Check to allow the action on all {{ data_type }} in your organization.</p>
+  <p>Check to allow the action on all {{ data_type }}s in your organization. This permission will be updated with new {{ data_type }}s as they are added.</p>
 </div>
 <div class="form-element full-width collapse show" id="object-selector">
   <div class="permissions-object-table">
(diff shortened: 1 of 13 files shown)


lboeman:
@wholmgren This should cover all of the issues in #173. I also corrected the aggregate messages because it only required one line of code.

sfa_dash/templates/forms/admin/permissions_form.html
```diff
@@ -37,7 +42,7 @@
 <div class="form-element full-width">
   <label>Applies to all</label>
   <input type="checkbox" name="applies-to-all" data-toggle="collapse" data-target="#object-selector">
-  <p>Check to allow the action on all {{ data_type }} in your organization.</p>
+  <p>Check to allow the action on all {{ data_type }}s in your organization. This permission will be updated with new {{ data_type }}s as they are added.</p>
```
wholmgren:
```suggestion
  <p>Check to allow the action on all {{ data_type }}s in your organization. This permission will also apply to all new {{ data_type }}s as they are added.</p>
```

PR #177, opened 2019-12-12
generalize table column filtering code and apply it to columns across…

### LINKED ISSUE #109: Expand table filters to other fields
expand the JS introduced in #107 to other fields, such as Variable

lboeman:
… dashboard
closes #109

### DIFF
--- a/sfa_dash/static/js/table-search.js
+++ b/sfa_dash/static/js/table-search.js
@@ -1,7 +1,7 @@
 /* Jquery for filtering and searching tables.
  * 
  * For simple column filtering, give an html table header an id of '<name>-header' and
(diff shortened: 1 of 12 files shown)


sfa_dash/static/js/table-search.js
```diff
@@ -1,8 +1,14 @@
 /* Jquery for filtering and searching tables.
+ * 
+ * For simple column filtering, give an html table header an id of '<name>-header' and
+ * give each td in it's column the '<name>-column' class. add a call to the block of
```
wholmgren:
```suggestion
 * give each td in its column the '<name>-column' class. add a call to the block of
```

Issue #170, opened 2019-11-26
404 after inactivity on /reports/


alorenzo175:
After a period of inactivity (i.e. access token has probably expired), if I go directly to the /reports/ page, I get a 404. If I instead go to /sites/ first, /reports/ works fine. I think this is related to how we get the token in api_interface/reports.py. Other interfaces, like sites, use the oauth_request_session which automatically requests a new access token if the current one is expired. I don't think 
```python
token = oauth_request_session.token['access_token'] 
```
does that. Probably need to check if the token is expired, and refresh if it is like https://requests-oauthlib.readthedocs.io/en/latest/oauth2_workflow.html#first-define-try-catch-tokenexpirederror-on-each-request

alorenzo175:
Actually, what probably makes the most sense (so that tokens are handled properly including stored) is to make a factory function for core.io.api.APISession where we can give it the OAuth2Session and still have the methods like list_reports etc.

lboeman:
This issue seems restricted to the dashboard, and a possibly introduces breaking change in core. Unless there's a benefit to having a token auto refresh feature in core, maybe the dashboard's report interface should be updated to work the same as the others (using oauth_request_session). Then it can load a data model object from the response where necessary?

PR #179, opened 2019-12-12
Report template

lboeman:
Opening this PR to have the code I've been working on for searching/sorting report metrics visible. The included `sfa_dash/templates/reptest.html` uses the metrics section from an early version of a report refactoring Tony is working on.


lboeman:
Adjusts `api_interface.reports` to build data model objects from the api responses instead of relying on core's APISession object. This should also close #170. and bring the dashboard up to date with beta4 reports.

sfa_dash/blueprints/reports.py
```diff
@@ -107,16 +110,9 @@ def post(self):
...
-                errors = {'error': ['Permission to create report denied.']}
-            else:
-                errors = {'error': ['An unrecoverable error occured.']}
-            return super().get(form_data=api_payload, errors=errors)
+            handle_response(reports.post_metadata(api_payload))
+        except DataRequestException as e:
+            return super().get(form_data=api_payload, errors=e.errors)
         return redirect(url_for(
             'data_dashboard.reports',
             messages={'creation': 'successful'}))
```
alorenzo175: redirect to `/reports/<uuid>`

sfa_dash/blueprints/reports.py
```diff
@@ -125,36 +121,113 @@ def post(self):
...
+        report_period = (end - start)
+        total_data_points = 0
+        raw_report = self.metadata['raw_report']
+        pfxobs = raw_report['processed_forecasts_observations']
+        for fxobs in pfxobs:
+            interval_length = pd.to_timedelta(fxobs['interval_length'],
+                                              unit='m')
+            fxobs_data_points = report_period / interval_length * 2
+            total_data_points = total_data_points + fxobs_data_points
+        return total_data_points < current_app.config['REPORT_DATA_LIMIT']
```
alorenzo175: so this could benefit from a count of the total valid points for each fx/obs stored in PreprocessingResult?

sfa_dash/blueprints/reports.py
```diff
@@ -125,36 +121,113 @@ def post(self):
...
+                    obs_list.append(observation)
+                aggregate['observations'] = obs_list
+                pair['aggregate'] = aggregate
+            forecast = handle_response(
+                forecasts.get_metadata(fxobs['forecast']))
+            forecast['site'] = handle_response(
+                sites.get_metadata(forecast['site_id']))
+            pair['forecast'] = forecast
+            pairs.append(pair)
+        metadata['report_parameters']['object_pairs'] = pairs
```
alorenzo175: getting the actual object pairs is unnecessary to render the report. You can just remove this loop and set object pairs to an empty list

sfa_dash/blueprints/reports.py
alorenzo175: and maybe any logic like this should go in core templates.py anyway? I don't think it's ever really useful to put the plots in if browser rendering will take ages.

lboeman: I think that leaving it up to the end user is somewhat valuable, and if we move this logic to core it should have some override flag. The current warning explains why we've omitted the timeseries, but If someone really wants to looks at a bokeh graph for a long period, I think we should allow them to download it.

lboeman:
After talking to @wholmgren, trying to squeeze in an interface for quality flag filters into this PR seems like a good idea. It would allow us to cover all of the functionality added in  corresponding core release. From looking at the datamodel in beta4,  a checklist of quality flags to exclude should work (similar to the checklist of included metrics). @alorenzo175 do you have any reservations about adding quality flag support here?

alorenzo175:
I might do it in a separate PR based on this one, just so we can get this merged and start testing it in dev

sfa_dash/blueprints/reports.py
```diff
@@ -174,34 +168,7 @@ def set_metadata(self, uuid):
...
-                pair['aggregate'] = aggregate
-            forecast = handle_response(
-                forecasts.get_metadata(fxobs['forecast']))
-            forecast['site'] = handle_response(
-                sites.get_metadata(forecast['site_id']))
-            pair['forecast'] = forecast
-            pairs.append(pair)
-        metadata['report_parameters']['object_pairs'] = pairs
+        metadata['report_parameters']['object_pairs'] = []
         self.metadata = metadata
```
alorenzo175: either here or in template_args we'll have to load it into the report object

alorenzo175:
We should use ``report['raw_report']['plots']['bokeh_version']`` in head with some default like 1.4.0 instead of the current fixed value.

alorenzo175:
I'm having trouble with the javascript ordering. Does it work for you?

lboeman:
It does work for me. Any js errors and what browser are you using?

alorenzo175:
firefox 71.0 on fedora. 
image

lboeman:
Ahh thanks. carets for the lose.

lboeman:
Okay. This should handle carets in the `data-metric` attribute now. From [the metrics page](https://solarforecastarbiter.org/metrics/), it doesn't look like there are any other characters that should cause problems.

alorenzo175:
yes, it works now. A few other things:

- Forecast: all should only have the total category and the Forecast: <indv> should have the other categories and not total 

image

- can you use the datamodel ALLOWED_METRICS and ALLOWED_CATEGORIES to map the names? perhaps by passing the dicts as json to the javascript

lboeman:
Those issues should be fixed now!
 I also caught the inverse issue where Category->Forecast nesting where all of the forecasts were showing up in 'total' and the 'all' forecast was showing up in other categories. So that should be fixed too.

sfa_dash/blueprints/main.py
```diff
@@ -312,3 +318,4 @@ class TrialsView(DataDashView):
 data_dash_blp.add_url_rule(
     '/aggregates/<uuid>/delete',
     view_func=DeleteAggregateView.as_view('delete_aggregate'))
+data_dash_blp.add_url_rule('/reptest/', view_func=RepTest.as_view('reptest'))
```
alorenzo175: don't need this anymore do we?

lboeman: ahh no, thanks.

lboeman:
> If I search for "all" the categories are screwy

Just as in, there are a lot of empty buttons that don't expand into anything?

alorenzo175:
yes 
image

alorenzo175:
image

alorenzo175:
this is odd
image


Maybe some of these are the expected behaviour?

lboeman:
Yeah, it's a tricky  fix, but I'll see what I can do. The issue is that it's really simple to select the plots themselves with the `data-forecast` attribute, but difficult to traverse the dom to figure out which of the containers are empty and which headers should also be hidden.

lboeman:
Alright, this *should* solve all of the issues listed above. A new function that is called on sorting or search update hides every container & header button combination that doesn't have any currently visible plots. 
e.g. given the sorting order category > forecast, all forecasts except for 'all' will be hidden in the 'Total' category because none of the forecast sub-categories will have plots.

PR #182, opened 2019-12-16
update run length/issue frequency

### LINKED ISSUE #66: run length / issue frequency
We should consider changing the label to just run length to avoid confusion.

lboeman:
closes #66


lboeman:
Tests are in fact passing although github has not updated. @wholmgren could you give this a quick?

wholmgren:
Let's wait on this. I don't know if it's actually a change for the better.

PR #184, opened 2019-12-16
Add download link to fx/cdf/obs pages

### LINKED ISSUE #144: persist datetime selections from Obs/Fx pages into download pages
So you don't have to fill out the fields twice. Another idea is to remove the download pages and move those options below the graph.

### LINKED ISSUE #157: Prepopulate time selection fields with defaults
Pages containing plots and start-end selection should prepopulate the input with the default (last three days) to put error messages in context.

### LINKED ISSUE #163: download data pages should link back to observation/forecast and site/aggregate
At least make the metadata table contain links instead of plain text names. Related: the site id is displayed instead of the site name.

Consider also adding something like /Site/Observations/Observation_Name/Download with links to the top of the frame.

lboeman:
Removes dedicated download pages in favor of a "Download Data" link next to the "Set Time Range" button, so users can view data before they download it.
closes #163 
closes #144 
closes #157 
Gives downloaded files a meaningful name, `<object name>_<start>_<end>.<type>` where type is 'csv' or 'json'.


sfa_dash/blueprints/aggregates.py
```diff
@@ -331,8 +337,60 @@ def get(self, uuid):
...
+        end_dt = pd.Timestamp(form_data['end'], tz='utc')
+        params = {
+            'start': start_dt.isoformat(),
+            'end': end_dt.isoformat(),
+        }
+        headers = {'Accept': form_data['format']}
+        return headers, params
+
+    def post(self, uuid):
+        form_data = request.form
```
alorenzo175: can format_params and post be shared with main.py?

lboeman: I think that would be ideal. The only obstacle I see is handling the routing, because right now the forms are set to just post too whatever the current path is, but that may be simple to handle.

lboeman:
I tried cleaning up some of this by moving some of the response logic to utils, and pushing `format_params`->`format_download_params` up to the parent class so it can be shared. The post endpoint could *maybe* just go in the base class as well.

I wasn't able to test this yet because there is an import error with my current version of core. I'll look into the issue further tomorrow.

alorenzo175:
Did you figure out the import error? how does this do with core master?

lboeman:
> Did you figure out the import error? how does this do with core master?

I'll rebase it upon #209 and make sure it's still functioning.

lboeman:
Working well with the new core. The added interfact looks like the screenshot here. The information buttons next to the format radio buttons link out to the format documentation on solarforecastarbiter.org
Screenshot from 2020-02-03 15-42-36

PR #185, opened 2019-12-19
fix report head spacing, give extra parameters text area a name

### LINKED ISSUE #183: inconsistent top margin/padding for left menu
Viewing a report shifts the left menu down slightly (~20 px). I'm guessing the report content also starts a little lower than the other site content.

lboeman:
closes #183 
This was a maddening and bizarre issue. Also fixes a problem introduced in #167, wherein I created a macro for the `extra_parameter` fields but omitted a name, causing a bad request key error on form submission.

### DIFF
--- a/sfa_dash/templates/data/aggregate.html
+++ b/sfa_dash/templates/data/aggregate.html
@@ -50,7 +50,9 @@
 <div class="row data-plots-wrapper">
 {{ plot | safe }}
 </div>
+{% if bokeh_script is defined %}
 {{ bokeh_script | safe }}
+{% endif %}
 <script src="/static/js/plot-bound-parsing.js"></script>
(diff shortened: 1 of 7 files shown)


sfa_dash/templates/data/aggregate.html
```diff
@@ -50,6 +50,7 @@ <h3 class="mt-3">Observations</h3>
 <div class="row data-plots-wrapper">
 {{ plot | safe }}
 </div>
+{{ bokeh_script | safe }}
```
wholmgren: do we need 

```
{% if bokeh_script is defined %}
{{ bokeh_script | safe }}
{% endif %}
```

here?

lboeman: ahh yes, thanks.

PR #192, opened 2020-02-05
Quality flags

lboeman:
Adds quality flags as a filter to the report form. In the future, this may be a sub-menu of some overarching 'filters' section.
Also adds quality flag filters and categories to the metadata displayed on the report deletion page.

### DIFF
--- a/sfa_dash/templates/forms/report_form.html
+++ b/sfa_dash/templates/forms/report_form.html
@@ -64,9 +64,7 @@
        <label>Quality Flag Filters</label><br>
        <p>Check quality flags to exclude from analysis. </p>
        {% for flag in quality_flags.keys() %}
-       {% if metric != 's' %}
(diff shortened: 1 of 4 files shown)


lboeman:
Form addition looks like this: 
Screenshot from 2020-02-06 12-18-09

Currently hardcoded the two supported quality flags in 'template_globals, but retained the dictionary structure from core so that it should be a simple drop in replacement when we need to expand (or not).

sfa_dash/templates/forms/report_form.html
```diff
@@ -91,6 +60,13 @@ <h5>Observation, Forecast pairs</h5>
      <input type="checkbox" name="categories" value="{{ category }}" {% if form_data %}{% if category in form_data['report_parameters']['categories'] %}checked{% endif %}{% else %}checked{% endif %}> {{ label }}<br/>
      {% endfor %}
      </div>
+     <div class="form-element">
+       <label>Quality Flag Filters</label><br>
+       <p>Checked quality flags are excluded from analysis. </p>
```
alorenzo175:
```suggestion
       <p>Data flagged with any checked quality flags is excluded from analysis. </p>
```

PR #195, opened 2020-02-06
include most things in manifest

alorenzo175:


alorenzo175:
will hold off until tests are also updated to be skipped

PR #205, opened 2020-03-03
B5 layout style

lboeman:
Updates & bug fixes to the ui and style for Beta 5. Also removing old and unused templates.


lboeman:
The aggregate overlap issue should be fixed now. The effective from/until fields should no longer contain a 'T' character, and will now include a 'Z' instead of '+0000' offset. This should help to both avoid wrapping and, in case of wrapping break on the date/time boundary and look like this: 
Screenshot from 2020-03-03 15-07-30

PR #207, opened 2020-03-09
Simplify use of handle_response

lboeman:
Rather than repeating `handle_response` everywhere, this moves it to wrap the `oauth_request_session` calls in the API interface.

### DIFF
--- a/sfa_dash/api_interface/util.py
+++ b/sfa_dash/api_interface/util.py
@@ -1,0 +1,73 @@
+from flask import request
+
+from sfa_dash.errors import DataRequestException
+
+
+# Utility functions pertaining to interactions with the api.
+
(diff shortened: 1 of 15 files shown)


sfa_dash/api_interface/util.py
```diff
@@ -0,0 +1,73 @@
...
+        If the status code received from the API could not be
+        handled.
+    """
+    if not request_object.ok:
+        errors = {}
+        if request_object.status_code == 400:
+            errors = request_object.json()
+        elif request_object.status_code == 401:
+            errors = {
+                '401': "You do not have permission to create this resource."
```
alorenzo175: is 401 only returned on a failed create?

PR #208, opened 2020-03-13
add logic to load plots only when shown in metrics section

lboeman:
Supports https://github.com/SolarArbiter/solarforecastarbiter-core/pull/359
Adds a function to load metric plots when they become visible. This function is called whenever sorting or searching happens, or when a metric section is expanded. Currently, loads plots as they become visible and does not remove them. 

This does introduce a slight delay after expanding a section before the plots load, which does increase to a second or two with a large number of plots showing. From testing with ~147 plots (5 fx, 7 metrics, 5 categories) expanding all of the metrics plots does not bog down the browser into being unresponsive.

### DIFF
--- a/sfa_dash/templates/head.html
+++ b/sfa_dash/templates/head.html
@@ -37,6 +37,8 @@
 <script src="/static/js/table-search.js"></script>
 <script src="/static/js/table-checkall.js"></script>
 
+<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
+
 {# inject the page_data json variable if provided by jinja #}
 {% if page_data is defined %}
 <script>
--- a/sfa_dash/templates/head.html
+++ b/sfa_dash/templates/head.html
@@ -37,7 +37,7 @@
 <script src="/static/js/table-search.js"></script>
 <script src="/static/js/table-checkall.js"></script>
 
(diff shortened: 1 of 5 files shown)


wholmgren:
sweet. do you know how to easily add a little spinning icon when something is loading? if not, don't worry about it.

lboeman:
Spinny loady doodads created!

sfa_dash/templates/head.html
```diff
@@ -37,6 +37,8 @@
 <script src="/static/js/table-search.js"></script>
 <script src="/static/js/table-checkall.js"></script>
 
+<script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
```
wholmgren: latest? seems safer to pin. only a matter of time before COVID-19 finds its way into the filthy javascript ecosystem.

lboeman: ahh yep, thanks for that! I added that in to start testing before plotly version was available.

lboeman: Wash your imports!

lboeman: And unfortunately Plotly doesn't have the same versions across languages, so we're going to have to keep an eye on plotly :[

lboeman:
@wholmgren any concerns before this is merged? The core counterpart was just merged.

PR #209, opened 2020-03-23
Hide actions user does not have permission to perform

lboeman:
Uses the `actions-on` endpoint added in https://github.com/SolarArbiter/solarforecastarbiter-api/pull/221 to provide a list of allowable actions for the current object and expose the `is_allowed` function to jinja templates. These actions are then used to hide some dashboard functionality. This works nicely for single object pages, e.g. A site, or observation.

There are a couple of shortcomings that It would be nice to remedy:
- Create permissions are only related to an object type, and cannot be queried by uuid as the endpoint I added is set up. Maybe a list of object types the user is capable of creating is always returned from these endpoints?

- Listing pages where delete or update buttons have been addded to the row for convenience are not hidden. For example, the reports listing page has a delete button for each report in the table. Querying for each of the listed objects is expensive, so a bulk actions query endpoint may be in order if it does not turn out to be terribly inefficient in mysql.

### DIFF
--- a/sfa_dash/templates/data/time_widgets.html
+++ b/sfa_dash/templates/data/time_widgets.html
@@ -1,4 +1,6 @@
 {% import "forms/form_macros.jinja" as form %}
+
+{% if is_allowed('read_values') %}
 {# Time widgets are used for either setting the bounds of the plot or download #}
 <div class="form-group mt-3">
 <div class="form-element">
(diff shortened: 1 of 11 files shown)


sfa_dash/templates/data/time_widgets.html
```diff
@@ -15,11 +15,13 @@
 <div class="form-element full-width">
   <button type="submit" form="plot-range-adjust" value="Submit" class="btn btn-primary">Update graph</button>
   <form action="" method="post" id="download-form" onsubmit="ParseStartEnd()">
+  {% if is_allowed('read_values') %}
```
alorenzo175: is there any point in adding any of the time widgets if the user can't read values and I assume there is no/an empty graph?

PR #220, opened 2020-04-06
Support reference forecasts, uncertainty, unit consistency

### LINKED ISSUE #214: add reference forecast option to report form
The user interface and/or API is the most challenging part of this. I'm sure core can adapt to whatever we need.

I think the core calculator code assumes a single reference forecast for all objects, so we should take that approach here too. 

We'll need to restrict reports to a single variable in order for this to make sense. 

Maybe `'reference_forecast'` is a new kind of key in `report_objects`?

@lboeman how much effort do you think this is going to be?

### LINKED ISSUE #217: enforce unit consistency in reports
The core datamodel checks for unit consistency for forecast/observations in a report. Maybe we should go further and ensure the variable is the same instead of just the units, but either way the report form should validate for these inconsistencies so we don't get a failed report

### LINKED ISSUE #218: option to account for uncertainty in report
https://github.com/SolarArbiter/solarforecastarbiter-core/pull/378 adds a capability for core to account for uncertainty in reports. This needs to be exposed on the dashboard too.

One suggestion is radio buttons for:

* Ignore uncertainty --> {'deadband': null}
* Set deadband to observation uncertainties --> {'deadband': 'observation_uncertainty'}
* Set deadband for all observations: --> {'deadband': input box value}

These would live near the metrics check boxes. 

Another option is to put the selection boxes within the "Observation, Forecast pairs" section and allow the user to set the deadband for each pair. That seems a little ambitious for now, though.

What do you think of feasibility @lboeman?

lboeman:
closes #218, closes #217, closes #214 
Features updated/ added in this PR:

Updates the list of pairs to display uncertainty and reference forecast data as shown below. The first pair displays an unset uncertainty. The second pair displays an unset reference forecast, and the result of "observation_uncertainty"(note that the value transmitted to the api is the string "observation_uncertainty" , but I've displayed the value for the user here). The third pair displays user provided uncertainty. 
Screenshot from 2020-04-06 15-11-52

per #217: When the first object pair is set, a units variable is set, and the available variables are limitted. 

per: #214 Adds a reference forecast selection area that is dependent on the forecast selection. This is limited to forecasts at the same site and variable and is required to have the same or shorter interval length. This results in listing most other forecasts at the same site.

Screenshot from 2020-04-06 15-18-20

per #218: Adds a new widget at the end of the object pair selection area to select the source of uncertainty per pair. 
Screenshot from 2020-04-06 15-20-41

Added a check when an object pair is added to uncheck and disable the "skill" metric and display a message if a reference forecast was not supplied. 
Screenshot from 2020-04-06 15-22-24

@wholmgren Do these seem in line with what you were thinking for the updated ui?

### DIFF
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -75,12 +75,12 @@ class ReportForm(BaseView):
 
         reference_forecasts = filter_form_fields('reference-forecast-',
                                                  form_data)
+
         uncertainty_values = filter_form_fields('deadband-value-', form_data)
         pairs = [{'forecast': f,
                   truth_types[i]: truth_ids[i],
                   'reference_forecast': reference_forecasts[i],
-                  'uncertainty': uncertainty_values[i],
-                 }
+                  'uncertainty': uncertainty_values[i]}
                  for i, f in enumerate(fx)]
         return pairs
 
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -72,7 +72,15 @@ class ReportForm(BaseView):
         # appropriately indexed forecast.
         truth_ids = filter_form_fields('truth-id-', form_data)
         truth_types = filter_form_fields('truth-type-', form_data)
-        pairs = [{'forecast': f, truth_types[i]: truth_ids[i]}
+
+        reference_forecasts = filter_form_fields('reference-forecast-',
+                                                 form_data)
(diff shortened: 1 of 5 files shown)


wholmgren:
looks great!

lboeman:
I think this is in a good spot, I will hold off to test against https://github.com/SolarArbiter/solarforecastarbiter-core/pull/378 to make sure nothing catches fire before merging.

PR #223, opened 2020-04-16
Event report forms

### LINKED ISSUE #216: support event forecast reports
less complicated version of #215 

core PR: https://github.com/SolarArbiter/solarforecastarbiter-core/pull/361

lboeman:
closes #216. 

Based on #220 

Adds a tabbed "report type" selector to to the top of the report form, to select between deterministic, event and probabilistic reports. Screenshot of event report form below. 

localhost_8080_reports_event_create

### DIFF
--- a/sfa_dash/blueprints/form.py
+++ b/sfa_dash/blueprints/form.py
@@ -459,8 +459,14 @@ forms_blp.add_url_rule('/forecasts/single/<uuid>/upload',
 forms_blp.add_url_rule('/forecasts/cdf/single/<uuid>/upload',
                        view_func=UploadForm.as_view('upload_cdf_forecast_data',
                                                     data_type='cdf_forecast'))
-forms_blp.add_url_rule('/reports/create',
-                       view_func=ReportForm.as_view('create_report'))
+forms_blp.add_url_rule('/reports/deterministic/create',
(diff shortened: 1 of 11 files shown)


lboeman:
This is pretty close to ready. I just wanted to double check if a user should be able to pair an event forecast with an aggregate, or should that selection be removed from the event report form?

wholmgren:
We're not supporting aggregate events.

alorenzo175:
Not sure if this should be in a separate PR, but the section headings don't support event metrics

image

lboeman:
> Not sure if this should be in a separate PR, but the section headings don't support event metrics
> 
> image

Is that the current version of this pr? should be fixed in https://github.com/SolarArbiter/solarforecastarbiter_dashboard/pull/223/commits/3bc552511c39723e369ad4fae4d39b22be60dfed

lboeman:
Last question (I think): Is there a set of default event metrics that should be selected? 

My test event report renders its metric section like this: 
Screenshot from 2020-04-22 15-38-17

And here is the current version of the form. I think It's ready sans default event metrics.
localhost_8080_reports_event_create (1)

wholmgren:
They work better in combination and there's not too many of them, so probably all of them.

Issue #226, opened 2020-04-23
make "Day of the week" unselected by default


wholmgren:
The "Day of the week" category is rarely interesting for applications other than net load. Let's disable it by default. I imagine it's possible to magically enable it if a net load forecast is selected.

PR #228, opened 2020-04-23
Clearsky and default categories

### LINKED ISSUE #225: add cloudy/clearsky flag to report options
https://github.com/SolarArbiter/solarforecastarbiter-core/pull/382 improved support for the cloudy/clearsky flag in core. I believe we're now ready to support it in reports. Is that right @alorenzo175 ? If so, @lboeman we need to expose it here on the dashboard. I'm guessing/hoping that's a quick fix that can make it into beta5.

lboeman:
Closes #225 and #226


sfa_dash/template_globals.py
```diff
@@ -30,14 +30,17 @@
...
 ALL_METRICS.update(ALLOWED_DETERMINISTIC_METRICS)
 ALL_METRICS.update(ALLOWED_EVENT_METRICS)
 ALL_METRICS.update(ALLOWED_PROBABILISTIC_METRICS)
 
 ALLOWED_QUALITY_FLAGS = {
-    'USER FLAGGED': 1,
-    'NIGHTTIME': 16,
+    'USER FLAGGED': 'USER FLAGGED',
+    'NIGHTTIME': 'NIGHTTIME',
+    'CLEARSKY EXCEEDED': 'CLEARSKY EXCEEDED (GHI ONLY)',
```
wholmgren, line 43: CLEARSKY not CLEARSKY EXCEEDED

wholmgren: Also just capitalize the O of Only

PR #238, opened 2020-04-29
Probabilistic report form

lboeman:
Creates a probabilistic report form that is accessible via the tabs introduced in #223. This will require updates to the API report schema for object pairs that I will add in a forthcoming PR as well as the datamodel updates in https://github.com/SolarArbiter/solarforecastarbiter-core/pull/426

At opening this PR the javascript handling of the probabilistic report form has been implemented. The dashboard is not configured to correctly parse submissions from this form.

Here's a screenshot of the current form. Nitpicks are welcome.
 
localhost_8080_reports_probabilistic_create

TODOs before merging:

- [x] Implement logic at the report view for accepting and formatting the API payload correctly. 
- [ ] Test against API. 
- [x] Clean up any remnants of the mockup process.
- [x] Update "Select constant values" prompt.


sfa_dash/templates/forms/probabilistic_report_form.html
```diff
@@ -0,0 +1,79 @@
...
+       <span class="period-end-help-text form-text text-muted help-text collapse" aria-hidden>
+           The end of the analysis period as an ISO 8601 datetime.
+       </span>
+     </div>
+     {# The following empty list is a target for the javascript found in report-handling.js
+        list items are added dynamically on page load #}
+     <h5>Observation, Probabilistic Forecast Pairs</h5>
+     <p>For each Observation, Probabilistic forecast pair selected
+        <a href="https://solarforecastarbiter.org/metrics/#crps">CRPS</a> will
+        be calculated.</p>
```
wholmgren, line 42: is this old text? I thought it should be something like "For each Obsevation, Forecast pair selected, the appropriate metrics will be calculated"

wholmgren: Also wondering about the name **Select Constant Values** - seems out of date with current intention (as I understood it)

lboeman: Yeah. There is a better label and explanation for the set of options that now includes the full cdf_forecast_group and its constant values, though I'm not coming up with anything quickly.
 I'm also just realizing we should probably relabel all of these **Forecast evaluation pairs** to be inline with the original Data Model document.

lboeman:
Today's team call brought up a couple of things to incorporate:

- Clearer description of constant value in the stored object pairs. e.g. instead of just having `<forecast name> <constant value> <units>` have a separate field with an explanation in line with Something like "Prob(x) < 2 W/m^2" and "Prob(x) = 10%"? .

- Potentially add a pop out box with full metadata for the pair's related objects.

The latter seems like a nice-to-have, but I think it should be a pretty quick and easy thing to implement if everyone sees value in it.

lboeman:
Updated the object pair boxes with a constant value field following the suggestion @dplarson brought up.
Also added a separate metrics section for groups and constant values.
 Screenshots are below.
Constant percentiles:
Screenshot from 2020-04-30 16-21-15

Constant variable values:
Screenshot from 2020-04-30 16-25-53

Metric sections:
Note that the CRPS field is disabled and is still checked automagically. When the form is submitted, the validation callback enables it so it is submitted.
Screenshot from 2020-04-30 16-36-28

dplarson:
@lboeman I definitely like the overall direction of the revisions. But I made a couple minor mistakes in my suggestion for the constant variable values text formatting:
1. it should be "less than or equal" to match how the probabilistic metrics are defined  (`o = np.where(obs <= fx, 1.0, 0.0)  # o=0: did not happen, o=1: did happen`)
2. it should the "less than or equal" statement should be inside the `Prob` operator

So it goes from `Prob(x) < 2 W/m^2` to `Prob(x <= 2 W/m^2)`, while `Prob(x) = 5%` stays the same.

**NOTE:** let me think about this a little more. I *think* my suggestions are correct, but I may be making a subtle mistake in my notation and/or there might be a clearer way to word all of this.

dplarson:
Maybe it would be clearer as:
- `Prob(obs <= fx) = 2%`: forecast a physical value (`fx`) with probability of 2%
- `Prob(obs <= 5 W/m^2)`: forecast a probability (`fx_prob`) for the observation being less than or equal to 5 W/m^2

wholmgren:
Nice work @lboeman and good suggestions @dplarson. 

A few additional things... 

I'd like to rethink is the `Constant value: Full probabilistic forecast group` labeling. A few ideas for how to do this differently for `ProbabilisticForecast`s:

1. Change value to values and list all of them. This would get ugly if there are many. Reject.
2. Remove the constant values line and don't replace it
3. Replace "constant values" with "Distribution" and don't say anything more
4. Replace "constant values" with "Distribution" and put something like `Prob(obs <= fx)` after it

Related: change the metrics subsection "Full Probabilistic Forecast Group Metrics" to "Distribution Metrics". Maybe also change "Constant Value Metrics" to "Binary Metrics". My goal is to try to use standard terminology and avoid our jargon where possible.

lboeman:
> Nice work @lboeman and good suggestions @dplarson.
> 
> A few additional things...
> 
> I'd like to rethink is the `Constant value: Full probabilistic forecast group` labeling. A few ideas for how to do this differently for `ProbabilisticForecast`s:
> 
> 1. Change value to values and list all of them. This would get ugly if there are many. Reject.
> 2. Remove the constant values line and don't replace it
> 3. Replace "constant values" with "Distribution" and don't say anything more
> 4. Replace "constant values" with "Distribution" and put something like `Prob(obs <= fx)` after it
> 
> Related: change the metrics subsection "Full Probabilistic Forecast Group Metrics" to "Distribution Metrics". Maybe also change "Constant Value Metrics" to "Binary Metrics". My goal is to try to use standard terminology and avoid our jargon where possible.

Does it also follow that the option in the select list should be changed to "Distribution", and this should also be reflected in the instructions with something like "select _Distribution_ to calculate the crps metric for this forecast"?

wholmgren:
Yes. I'd like to add an adjective or some phrase to make it more clear. The problem with "full" is that it could imply that the number of constant values is sufficient to "fully" specify the distribution - whatever that means - and that is not always going to be the case.

lboeman:
Alright here's a new full screenshot of the form. I think I've captured most of your comments @wholmgren. 
distribution

wholmgren:
@dplarson @alorenzo175 I've been discussing how to improve this with @lboeman and looking for your feedback on this proposal:

1. Move and reword the descriptive text:

The Solar Forecast Arbiter supports the specification of probabilistic forecasts in terms of a cumulative distribution function (CDF). First select, the distribution of interest. Second, select the portion of the distribution to be evaluated. Select Distribution to calculate the CRPS metric for the entire forecast distribution. Select one or more forecasts for portions of the distribution to calculate binary metrics for each forecast. Multiple forecasts may be selected using ctrl+click for Windows or command+click for Mac.     
`<PROBABILISTIC FORECAST SELECT BOX>`
`<DISTRIBUTION + CONSTANT VALUES SELECT BOX>`

2. The `<DISTRIBUTION + CONSTANT VALUES SELECT BOX>` will use the label style that @dplarson proposed e.g. `Prob(obs <= 5 W/m^2)`

3. The selected Obs/Fx pairs metadata boxes will go back to appending the additional metadata to name, but using @dplarson's scheme e.g. `1 hour test forecast (Distribution)` and `1 hour test forecast (Prob(obs <= 5 W/m^2))`

We thought it was more clear overall (certainly not perfect) and also gets us away from the constant values jargon.

lboeman:
I've realized I've missed reference forecasts here. While you can select a reference ProbabilisticForecasts, there is no support for pairing constant values together. 

Is the expectation that a user can only select a reference forecast with the same constant_value field? We may need to remove the ability to select multiple forecasts so that we can filter potential reference forecasts correctly, and make the selection process more obvious.

wholmgren:
Correct.

dplarson:
I think the proposal makes sense. The only thing I'm not 100% sure on is whether there is a "better" term than `binary` to describe the non-distribution metrics, but I have neither an alternative to suggest nor a clear argument for not using `binary`. So probably best to just ignore my comment and move forward with the proposal exactly as is.

alorenzo175:
I agree with 1 and 2. For 3, I think there could simplify further by having the distribution metadata box, and then the constant value/probability only as a little box indented under the distribution box. Something like 
```
Forecast: 1 hour forecast                                            REMOVE ALL
Observation:
...
Uncertainty: Ignore Uncertainty

    Distribution                                                               REMOVE
    Prob(obs < 5 W/m^2)                                                     REMOVE
    Prob(obs < 10 W/m^2)                                                   REMOVE
    ...

Forecast: next distribution
...
    Distribution
    All Probabilities
``` 

Just a suggestion and maybe a little unruly in the javascript, but once someone adds multiple forecasts and a handful of constant values for each, the boxes will be really hard to sort through

wholmgren:
@alorenzo175 that's a great idea. @lboeman is it feasible?

wholmgren:
@dplarson thanks for the feedback. I see "dichotomous" used seemingly interchangeably with "binary". Preference?

lboeman:
> @alorenzo175 that's a great idea. @lboeman is it feasible?

Yes. I think that's a much more manageable approach. Thanks @alorenzo175.  

I had one more question regarding reference forecasts. When selecting a reference forecast to compare to a distribution, does the reference forecast need to have the same set of constant values as the forecast? Or can the forecast have a subset of the reference forecast's constant values.
e.g.
```
1. forecast[constant_values] == [ 1,2,3,4] => reference_forecast[constant_values] == [1,2,3,4]
OR
2. forecast[constant_values] == [1,2,3,4] =>  each  of [1,2,3,4] in reference_forecast[constant_values]
```

wholmgren:
We currently only support the Brier Skill Score so reference forecasts should only be able to be selected for `ProbabilisticForecastConstantValues` and the reference should be required to have the same properties (axis and constant_value). 

There is a CRPS Skill Score but we've not implemented it. If/when we do implement it, we probably should allow `ProbabilisticForecasts` with differing `constant_values` to be compared - presumably a poorly specified distribution (i.e. too few points) would result in a worse CRPS. I am not 100% sure about the statistics of this, though.

dplarson:
> @dplarson thanks for the feedback. I see "dichotomous" used seemingly interchangeably with "binary". Preference?

No preference, but thanks for asking.

lboeman:
I've updated the pair display to roughly what @alorenzo175 suggested. The pairs are now grouped by a combination of `ProbabalisticForecast`, `Observation` or `Aggregate` and the uncertainty. The constant values and their reference values and their reference forecasts are displayed as a list in an indented section.
Screenshot from 2020-05-06 13-46-52

wholmgren:
Great! 

Are the pairs displayed in the order they're selected? I wonder if that's good or if they should be ordered by name and then reference forecast.

What needs to happen to push this to completion?

lboeman:
> Great!
> 
> Are the pairs displayed in the order they're selected? I wonder if that's good or if they should be ordered by name and then reference forecast.
> 
> What needs to happen to push this to completion?

The pairs are displayed in the order their added. My feeling is that it makes more sense to see the last thing you added appended to the end of the list. But I'm not strongly opposed to sorting them.

I think this is very close to being finished. I'd like to test it against the api with a working core back end before it gets merged.

sfa_dash/static/js/probabilistic-report-handling.js
```diff
@@ -0,0 +1,1055 @@
...
+        'ghi': 'W/m^2',
+        'dni': 'W/m^2',
+        'dhi': 'W/m^2',
+        'poa_global': 'W/m^2',
+        'relative_humidity': '%',
+        'ac_power': 'MW',
+        'dc_power': 'MW',
+        'availability': '%',
+        'curtailment': 'MW',
+    }
```
wholmgren, line 64: should we put this in a shareable module?

sfa_dash/static/js/probabilistic-report-handling.js
```diff
@@ -0,0 +1,1055 @@
...
+            'ghi': 'GHI',
+            'dni': 'DNI',
+            'dhi': 'DHI',
+            'poa_global': 'Plane of Array Irradiance',
+            'relative_humidity': 'Relative Humidty',
+            'ac_power': 'AC Power',
+            'dc_power': 'DC Power',
+            'availability': 'Availability',
+            'curtailment': 'Curtailment'
+        }
```
wholmgren, line 136: another dict that looks useful in other places.

sfa_dash/static/js/probabilistic-report-handling.js
lboeman: I think there's a bunch of common functions that could be pulled out and organized into a js module, especiallly across the different report handling files. It looks like IE doesn't support modules though.

sfa_dash/templates/head.html
```diff
@@ -35,14 +35,22 @@
...
-    var MAX_PLOT_DATAPOINTS = {{ MAX_PLOT_DATAPOINTS }};
+    var sfa_dash_config = {
+      MAX_DATA_RANGE_DAYS: {{ MAX_DATA_RANGE_DAYS }},
+      MAX_PLOT_DATAPOINTS: {{ MAX_PLOT_DATAPOINTS }},
+      VARIABLE_NAMES: JSON.parse('{{ variable_names | tojson }}'),
+      VARIABLE_UNIT_MAP: JSON.parse('{{ variable_unit_map | tojson }}'),
+      METRIC_CATEGORIES: JSON.parse('{{ metric_categories | tojson }}'),
+      DETERMINISTIC_METRICS: JSON.parse('{{ deterministic_metrics | tojson }}'),
+      ALL_METRICS: JSON.parse('{{ all_metrics | tojson }}'),
+    }
```
lboeman, line 53: @wholmgren, I added a  global `sfa_dash_config` variable that gets loaded in the <head> element, and we can load python dicts into it with the `tojson` filter. What do you think of this? There are still common functions that could be pulled into a module, but this will help with the configuration we want to pull in from core.

wholmgren: sounds reasonable to me

lboeman:
> I'm ok with it but could we first hide the link to the form?

Yep, done!

alorenzo175:
good with me too

PR #245, opened 2020-05-14
Remove "created at" column from reports table

### LINKED ISSUE #242: dashboard shows first created time of updated reports
Dashboard report list shows a `Created` datetime of 2020-02-07

Screen Shot 2020-05-05 at 7 38 50 PM

Report metadata shows `Generated at` of 2020-05-05

Screen Shot 2020-05-05 at 7 39 03 PM

The language and times should be consistent. Do we need to display both times? I'd rather not.

lboeman:
closes #242

### DIFF
--- a/sfa_dash/static/css/styles.css
+++ b/sfa_dash/static/css/styles.css
@@ -378,6 +378,9 @@ a.help-button{
 .empty-reports-list{
     list-style: none;
 }
+table.table.reports-table{
+    table-layout: auto;
+}
 a:not([href]).object-pair-button{
     color: #007bff;
 }
--- a/sfa_dash/static/css/styles.css
+++ b/sfa_dash/static/css/styles.css
@@ -609,8 +609,9 @@ input#custom-deadband{
     border-radius: .5em;
     padding: .5em;
 }
-a.report-details-expander{
-    color: #007bff!important;
+div.report-details-expander{
+    width: fit-content;
+    text-align: center;
 }
 a.report-details-closer{
     position: absolute;
@@ -620,4 +621,7 @@ a.report-details-closer{
 }
 .report-metadata-popup ul.data-metadata-fields {
     padding-left: .5em;
+}
+.reports-table-row:nth-child(n+3) {
+    width: fit-content;
 }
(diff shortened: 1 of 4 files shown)


lboeman:
Removes the 'Created' column and removes the fixed table column widths such that report names can expand, reducing vertical table height. 
Screenshot from 2020-05-14 12-51-36

wholmgren:
I have conflicting opinions about this. 

Definitely like that there's no longer confusion between "created at" and "generated at". Like that there's more room for the name.

The created at column was useful when distinguishing my poorly named reports. Is "generated at" an option? Would we go so far as including start and end too, maybe in a pop up?

Blah. What do you think?

lboeman:
> I have conflicting opinions about this.
> 
> Definitely like that there's no longer confusion between "created at" and "generated at". Like that there's more room for the name.
> 
> The created at column was useful when distinguishing my poorly named reports. Is "generated at" an option? Would we go so far as including start and end too, maybe in a pop up?
> 
> Blah. What do you think?

I think it's probably useful to have a pop up so that users wont need to load the report. The "generated at" field is part of the `RawReport` object, so it's not loaded during the listing.

lboeman:
I've implemented a simple popup here. It's the same metadata being shown at the top of the delete report confirmation page.
Screenshot from 2020-05-15 15-46-39
What do you think @wholmgren ?

wholmgren:
Is it a hover over pop up? Or is there a subtle button I’m not seeing on my phone? Looks like a lot for a hover.

wholmgren:
Otherwise looks great!

lboeman:
> Otherwise looks great!

I've added a "Details" button next to status that causes the popup. Now looking at it, I see that since it matched the title text, it's not immediately apparent. I'll try a button.

lboeman:
Here's an updated screenshot with the details trigger as a button. Much less subtle.
Screenshot from 2020-05-18 09-18-15

wholmgren:
Yes much better. I wonder if it should be "Metadata details" or just "Metadata". "Details" could give the impression that you'll get the whole report. But I guess it's not a big deal since you don't leave the page so maybe best to keep it simple. Shrug.

lboeman:
> Yes much better. I wonder if it should be "Metadata details" or just "Metadata". "Details" could give the impression that you'll get the whole report. But I guess it's not a big deal since you don't leave the page so maybe best to keep it simple. Shrug.

I think "Metadata" gives a clearer idea of what to expect there, so I went ahead and updated it. Ok to merge?

PR #246, opened 2020-05-14
Add Recompute button to reports

### LINKED ISSUE #224: add recompute report button somewhere
perhaps at the top of the report page? Should probably also catch errors when trying to render the report here https://github.com/SolarArbiter/solarforecastarbiter_dashboard/blob/7f538d90331e27bc145a623aebef580738056462/sfa_dash/blueprints/reports.py#L163

lboeman:
closes #224 
Adds a recompute button to reports just below the report's title. Positioning seems a little bit awkward but I'm not sure there's a better place for it.
Screenshot from 2020-05-14 15-44-25

### DIFF
--- a/sfa_dash/api_interface/reports.py
+++ b/sfa_dash/api_interface/reports.py
@@ -19,3 +19,8 @@ def post_metadata(report_dict):
 def delete(report_id):
     delete = delete_request(f'/reports/{report_id}')
     return delete
+
+
+def recompute(report_id):
+    recompute = get_request(f'/reports/{report_id}/recompute')
+    return recompute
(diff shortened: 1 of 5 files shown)


alorenzo175:
as for the location, I think on the the right edge aligned with the title would be better maybe? but not if the title is long?idk, it's ok where it is

lboeman:
> as for the location, I think on the the right edge aligned with the title would be better maybe? but not if the title is long?idk, it's ok where it is

Yeah, just trying it out, it causes some weird wrapping for the header text. Where it is will work for now. Thanks for reviewing.

wholmgren:
I'm good with the simple below the title positioning. 

Is the button only visible if you actually have the ability to recreate the report (rather than just view it)?

lboeman:
> I'm good with the simple below the title positioning.
> 
> Is the button only visible if you actually have the ability to recreate the report (rather than just view it)?

It only shows up if you have "Update" permissions on the report. However, you need access to all of the objects referenced by the report for it to recompute successfullly.

lboeman:
Clear to merge this?

wholmgren:
Yes

PR #252, opened 2020-05-19
Display extra parameters in drop down

### LINKED ISSUE #204: Extra parameters viewing
Enable users to view extra parameters of metadata objects. Probable wrapped in `<pre>` tags.

Need to make sure we safely escape user input, but print extra parameters as provided.  e.g. since we're encouraging use of formatted input display json and yaml as provided by the user

lboeman:
closes #204 
Adds an expandable extra parameters field to the box of metadata fields. This was suprisingly simple, and Jinja does a good job of escaping the contents of the field. To be sure we don't accidentally include the extra parameters in a block with autoescape turned off, I've wrapped the print statement in an explicit `{% autocomplete true %}` block, and put all of this html in a template to be included elsewhere.
Screenshots:
Collapsed
Screenshot from 2020-05-19 11-21-51
Expanded with json:
Screenshot from 2020-05-19 11-22-15
Expanded with newlines and malicious extra parameters:
Screenshot from 2020-05-19 11-23-19

### DIFF
--- a/sfa_dash/static/css/styles.css
+++ b/sfa_dash/static/css/styles.css
@@ -381,13 +381,13 @@ a.help-button{
 table.table.reports-table{
     table-layout: auto;
 }
-a:not([href]).object-pair-button{
+a:not([href]).collapser-button{
     color: #007bff;
 }
-a:not([href]).object-pair-button:hover{
+a:not([href]).collapser-button:hover{
     color: #0056b3;
 }
-.object-pair-button::after{
+.collapser-button::after{
     width: 0; 
     height: 0; 
     border-left: 5px solid transparent;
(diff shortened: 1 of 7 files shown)


lboeman:
> Looks great! You’ve tested for all data model objects? Not sure from the diff what needs to be special cases and what happens automatically.

Yep! I've tested sites, observations, forecasts, cdf forecast groups, cdf forecast constant values and aggregates. 
Sites and aggregates are the only cases where there templates include the `extra_parameters.html` template directly because they don't extend the `data_metadata.html` template. Still kind of confusing because templates are organized and labelled with the weird pre-alpha terminology I devised. They could use better names.

wholmgren:
Ok merge when ready.

PR #258, opened 2020-05-22
Add validation for name field and uploaded data

### LINKED ISSUE #254: Clarification and/or validation of file size limits for uploads?
While trying to upload observation data for a new site using the dashboard, I ran into an "Internal Server Error" ~30-seconds after starting the upload of a ~9 MB CSV file. While this may just be a one-time error, it made me think that it may be worthwhile to do some testing with files of different sizes in order to confirm potential limitations for uploading data files and provide some documentation on the subject. For example, telling users to chunk data files if the total data is over X MB (e.g. `observation.csv` becomes `observation_1.csv`, `observation_2.csv`, etc.) or advising users to upload big data files via the API instead of the dashboard (if it turns out the API is more robust for large files).

Just a thought.

lboeman:
closes #254 Adds a validator for name fields that supplies a meaningful message on user input so users don't have to guess about allowed characters.
Screenshot from 2020-05-22 13-59-37

This also checks file sizes in the file upload form and limits them to 16MB displaying an error when the size was exceeded. 
The timeout issues are most influenced by the formatting and storage of data on the API. Regardless of original format, files containing ~200k data points were about the limit of stability (on dev), so I arrived at 16MB based on the average size of some randomly generated bad-case JSON files (where date format has resolution finer than a minute, float precision of 10 decimal places) of 200k points. CSV files were ~ half the size of the json files and upload quickly but the subsequent validation process involving a get, processing and post is where the timeout occurs for larger uploads.

 I've added a note near the file selector warning the users of this limitation and the rounding of data when stored. A warning will appear if the user selects a file that is too large.

Screenshot from 2020-05-22 13-56-44

### DIFF
--- a/sfa_dash/static/js/form-validation.js
+++ b/sfa_dash/static/js/form-validation.js
@@ -1,0 +1,46 @@
+MAX_FILESIZE = 16777216;
+$(document).ready(function(){
+    // Custom validity handlers to supply users with more meaningful messages
+    // when they provide invalid input.
+    if(document.getElementsByName('name').length >0){
+        // Listen for invalid input
+        document.getElementsByName('name')[0].addEventListener(
+            'invalid', function(){
+                this.setCustomValidity(
+                    `Name may only contain characters a-z, A-Z, 0-9,
+                    '(', ')', '-', and '_'.`);
+            }
+        );
+        // clear any invalid state and check for validity to trigger an invalid
+        // event
+        document.getElementsByName('name')[0].addEventListener(
+            'input',function(){
+                console.log('oninput fire');
+                this.setCustomValidity('');
+                this.checkValidity();    
+            }
+        );
+    
+    }
+
+    // Validate file uploads are within the required range
+    $('input:file').change(function(){
+        if(this.files.length > 0){
+            if(this.files[0].size > MAX_FILESIZE){
+                $(this).after(
+                    $('<div>')
+                        .addClass(['alert', 'alert-danger'])
+                        .html(
+                            `The file selected is too large. The maximum file
+                             upload file size is
+                             ${MAX_FILESIZE / (1024 * 1024)}MB. Your file is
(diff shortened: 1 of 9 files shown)


sfa_dash/static/js/form-validation.js
```diff
@@ -0,0 +1,45 @@
...
+                this.setCustomValidity(
+                    `Name may only contain characters a-z, A-Z, 0-9,
+                    (, ), -, and _.`);
+            }
+        );
+        // clear any invalid state and check for validity to trigger an invalid
+        // event
+        document.getElementsByName('name')[0].addEventListener(
+            'input',function(){
+                console.log('oninput fire');
```
alorenzo175, line 18: just for development?

sfa_dash/static/js/form-validation.js
```diff
@@ -0,0 +1,45 @@
+MAX_FILESIZE = 16777216;
+$(document).ready(function(){
+    // Custom validity handlers to supply users with more meaningful messages
+    // when they provide invalid input.
+    if(document.getElementsByName('name').length >0){
+        // Listen for invalid input
+        document.getElementsByName('name')[0].addEventListener(
+            'invalid', function(){
+                this.setCustomValidity(
+                    `Name may only contain characters a-z, A-Z, 0-9,
```
alorenzo175, line 10: looks like we're missing comma, apostrophe, and space? Might be easier to understand if they are all together without the commas like ``a-zA-Z0-9(), -'_``

alorenzo175:
oh and might need a rebase to get the tests running

sfa_dash/static/js/form-validation.js
lboeman: Yep, thanks for catching that.

lboeman:
I've only included the time range metadata on pages that include plotting, would it be useful to include 
 elsewhere? The only other places of interest would be the data upload and delete confirmation pages.

PR #262, opened 2020-05-26
Redirect to the report view on successful recompute

### LINKED ISSUE #261: redirect away from /recompute
when a user clicks the recompute button, they are directed to ``dashboard/reports/<report id>/recompute``. If one refreshes, another recompute request would be sent. Instead, should redirect to the main report page

lboeman:
To close #261. Redirects to the report view on a successful recompute, instead of just returning a response from the `/reports/<report_id>/recompute`. This still returns a response from the `/recompute` endpoint on an error, so that errors are displayed. @alorenzo175 any objection to storing/popping messages and errors from a session object? Then we could just set the messages and redirect instead of doing the return another functions result with arguments solution.

### DIFF
--- a/sfa_dash/blueprints/base.py
+++ b/sfa_dash/blueprints/base.py
@@ -1,7 +1,7 @@
 from collections import OrderedDict
 
 
-from flask import url_for, render_template, request, session
+from flask import url_for, render_template, request, flash
 from flask.views import MethodView
 import pandas as pd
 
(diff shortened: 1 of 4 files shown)


alorenzo175:
> any objection to storing/popping messages and errors from a session object?

nope

lboeman:
> This works, but what do you think of message flashing? Seems like it will also take care of making sure the messages are only available to the next request and not some stale message that wasn't popped earlier.
> 
> https://flask.palletsprojects.com/en/1.1.x/patterns/flashing/#message-flashing-pattern

Message flashing seems like a better solution. Just have to flatten any API schema errors.

lboeman:
I've adjusted to use message flashing for messages/errors/warnings. I'll make another PR that updates throughout the dashboard. Thanks for the suggestion, I guess I had the wrong impression of how flask's message flashing worked from a while ago.

sfa_dash/templates/sections/notifications.html
```diff
@@ -1,29 +1,38 @@
-{% if messages is defined %}
```
alorenzo175, line 1: Not critical, but it would be nice not to have the empty divs floating around. Maybe something like
```
{% set flashed_messages = get_flashed_messages(category_filter=['messages']) %}
{% set flashed_warnings ...%}
{% if messages is defined or (flash_messages | length) > 0 %}
<div class="messages">
  <ul class="message-list notifcation-list">
    {% if messages is defined %} # still need this for now
...
```

sfa_dash/blueprints/base.py
```diff
@@ -185,6 +185,19 @@ def get_breadcrumb_dict(self):
...
+    def flash_api_errors(self, errors):
+        """Formats a dictionary of api errors and flashes them to the user on
+        the next request.
+
+        Parameters
+        ----------
+        errors: dict
+            Dict of errors returned by the API.
+        """
+        to_flash = [f'({key}) msg.join(", ")' for key, msg in errors]
```
alorenzo175, line 197:
```suggestion
        to_flash = [f'({key}) {msg.join(", ")}' for key, msg in errors]
```
and do we need errors.items(). make sure to test this

PR #266, opened 2020-05-27
Correct API 404 handling improve testing of GET views

lboeman:


lboeman:
Working on adding tests to ensure that 404s are handled correctly to add to this PR.

lboeman:
Adds testing for how views handle receiving 404s when there is a single object and when the path requires two uuids and one of them isn't accessible or does not exist for the user.

There were a few bugs found that are also fixed in this PR:

- Pages for granting and  removing roles from a user were not migrated to expect a DataRequestException from its interface. This is fixed, and now if the role is not accessable, the user is redirected to the referring page and an error message is displayed.

- Forms for creating a new obs/fx/cdf at a site or aggregate failed if the site or aggregate did not exist. Now the user will be redirected to the listing page for whatever type of object they were trying to create (e.g. the Observations page).

- The page for a Probabilistic forecast would throw an error instead of displaying a 404, due to improper nesting of the finally block of a try/catch.

PR #271, opened 2020-06-02
Fix prob. fx metadata pages

alorenzo175:


sfa_dash/templates/data/aggregate.html
```diff
@@ -2,9 +2,6 @@
 {% block content %}
 {{ metadata_block | safe }}
 {% include "sections/notifications.html" %}
-{% if metadata is defined %}
```
lboeman, line 5: Thanks for catching this. Tests are failing because we still need this line here, since metadata is used below to build the table of aggregate observations. Otherwise looks good.

alorenzo175: I'll just revert actually and open an issue to clean up the templates

PR #275, opened 2020-06-02
Object cloning

### LINKED ISSUE #165: report from existing report, metadata from existing metadata
use an existing report to prepopulate a new report submission form

worth thinking about if the same code could apply to all of the metadata objects

lboeman:
Works towards #165 
Introduces a `form_utils` subpackage and `FormConverter` abstract base class with functions for converting from form submission to an api compliant dict and vice versa. This removes the form parsing code from the View code and makes it a little easier to follow. I have implemented converters for Sites, Observations, Forecasts, Probabilistic Forecasts and Aggregates. The aggregate converter only parses the base aggregate metadata and not it's corresponding observations. I have added stubs for a `ReportConverter` class, but held off on implementation. Digging into the reports code here will overlap with #238 significantly, and it would be easier to implement after that's merged.

I've also added `/clone` endpoints to copy existing metadata for all of the above metadata except for aggregates. I have not yet added links to these endpoints.


sfa_dash/blueprints/base.py
```diff
@@ -249,6 +250,75 @@ def _pop_nonjson(meta_dict):
             return new_dict
         return _pop_nonjson(self.metadata)
 
+    def set_site_or_aggregate_metadata(self):
+        """Searches for a site_id or aggregate_id  in self.metadata
+        and loads the expected metadata object from the api in either
+        the 'site' or 'aggregate' key. If the object could not be retrieved,
+        sets a warning, sets the 'plot' template argument to None and
```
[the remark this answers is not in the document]
lboeman: The timeseries plotting in core requires a full datamodel object, which we can't build if the user doesn't have access to the site/aggregate. :/

sfa_dash/blueprints/form.py
```diff
@@ -446,6 +278,50 @@ def post(self, uuid):
...
+                        return redirect(f'data_dashboard.{self.data_type}',
+                                        uuid=uuid)
+                    self.set_site_or_aggregate_link()
+
+                if 'site' in self.metadata:
+                    template_args['site_metadata'] = self.metadata['site']
+                    template_args['metadata'] = self.render_metadata_section(
+                        template_args['site_metadata'])
+                elif 'aggregate' in self.metadata:
+                    # maybe won't work.
```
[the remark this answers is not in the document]
lboeman: Thanks, I forgot to address this. I'm updating the forecast form such that this works, and I' opened #278 so I can clean up the separate aggregate endpoint in another PR as not to polute this one further.

sfa_dash/templates/forms/cdf_forecast_group_form.html
```diff
@@ -6,17 +6,22 @@
...
 {% endif %}
 {% block fields %}
-<input type="hidden" name="site_id" value="{{ site_metadata['site_id'] }}">
 <div class="form-element">
   {{ form.input('Name', 'text', 'name', 'Name for the Forecast',
                 required=True, form_data=form_data) }}
 </div>
+{# if a site is defined, allow users to select a variable, else fetch the
+   variable from aggregate metadata #}
+{% if site_metadata is defined %}
```
alorenzo175, line 15: is there a way site_metadata won't be defined. For example, if the user does not have access to the site metadata but tried to clone a forecast? And then a variable couldn't be set

sfa_dash/templates/forms/base/creation_form.html
```diff
@@ -7,7 +7,15 @@
 {{ metadata | safe }}
 {% include "sections/notifications.html" %}
 {% if form_title %}<h3>{{ form_title }}</h3> {% endif %}
-<form action="{% if data_type == 'site' %} {{ url_for('forms.create_' + data_type) }}{% else %}{{ url_for('forms.create_' + data_type, uuid=site_metadata['site_id']) }} {% endif %}" method="post" enctype='{{ form_enc }}' id="{{ data_type }}-form">
+{% if data_type in ['site', 'aggregate'] %}
+<form action="{{ url_for('forms.create_' + data_type) }}" method="post" enctype='{{ form_enc }}' id="{{ data_type }}-form">
+{% elif site_metadata is defined %}
+<form action="{{ url_for('forms.create_' + data_type, uuid=site_metadata['site_id']) }}" method="post" enctype='{{ form_enc }}' id="{{ data_type }}-form">
+  <input type="hidden" name="site_id" value="{{ site_metadata['site_id'] }}">
```
alorenzo175, line 14: similar to the other comment. Is there a way site_metadata is not defined, but the site is valid and one could still create a forecast/obs for it?

sfa_dash/templates/forms/cdf_forecast_group_form.html
lboeman: Users should be directed away from the form if the associated site/aggregate is inaccessible before the template is rendered. This is to enforce the rule that a user must have read permissions on a site/aggregate to create observations or forecasts on it. The situation where neither site_metadata or aggregate_metadata will not be set should only be when creating or cloning a site or aggregate.
This does make me realize I need to think about controlling the display of a clone button, which I haven't added yet, because simple read permissions will not suffice.

PR #277, opened 2020-06-04
Set default timerange to display forecasted data

### LINKED ISSUE #276: change default display times for forecast data
Follow up to #267... plot yesterday through all available forecast times instead of just last 3 days.

Screen shot of default view for forecasts I submitted for the test trial that had 7 days of data:

Screen Shot 2020-06-03 at 9 51 17 AM

I'd like for the default to be this instead:

Screen Shot 2020-06-03 at 9 52 16 AM

lboeman:
Closes #276 

If start and end query parameters are not present and future data exists, displays the future data with a start 1 day prior to the time of the request. Also handles a bug where setting the query parameters to something that parses as NaT got through the logic, causing an error.

### DIFF
--- a/sfa_dash/blueprints/base.py
+++ b/sfa_dash/blueprints/base.py
@@ -105,6 +105,7 @@ class BaseView(MethodView):
         """
         start_arg = request.args.get('start')
         end_arg = request.args.get('end')
+        utc_now = pd.Timestamp.utcnow()
         try:
             end = pd.Timestamp(end_arg)
         except ValueError:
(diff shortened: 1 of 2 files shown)


sfa_dash/blueprints/base.py
```diff
@@ -105,15 +105,11 @@ def parse_start_end_from_querystring(self):
         """
         start_arg = request.args.get('start')
         end_arg = request.args.get('end')
-        # if start and end are not present, set them to an invalid datetime
-        # to trigger value error
-        if not start_arg:
-            start_arg = "x"
-        if not end_arg:
-            end_arg = "x"
```
wholmgren, line 113: define `utc_now = pd.Timestamp.utcnow()` up here so you're using a consistent value throughout the function.

PR #282, opened 2020-06-09
Breadcrumb tuples

### LINKED ISSUE #280: Fix breadcrumb bug
Using dicts to store breadcrumb information is causing issues when a site and obs (or other metadata) have the same name. This can be remedied by using a list of 2-tuples, and that should also result in easier to follow code.

lboeman:
closes #280 
Updates breadcrumb creation to be based on a list of 2 tuples of `link text, url` to avoid avoid key collisions in the old dictionary design.


wholmgren:
Can we merge this and test on dev?

PR #283, opened 2020-06-10
Cleanup templates

lboeman:
Things handled in this PR:

- Removes unused templates.
- Refactors the common metadata, notifications and subnav section as well as loading the js `metadata` variable into the `dash/data` template.
- Refactors the `CreateForm` class to work with aggregates or sites, and removes the Aggregate specific classes. 
- Adds comments to clarify design and usage for areas touched in this PR.


lboeman:
I've patched up those two issues, and went through the Views and altered them to consistenty configure the `template_args` dictionary in the function `set_template_args`. The HTTP verb methods of the Views should now only really contain code for data get/post and handling the resulting errors, with a final call to `set_template_args` and then template rendering.

PR #285, opened 2020-06-17
Climate zones

### LINKED ISSUE #279: Integrate climate zones into dashboard
We can start by displaying the list of climate zones for each site on the metadata pages maybe after #275. Then I think we need to be able to filter sites based on the zone.

Maybe eventually we display a [leaflet](https://leafletjs.com/examples/geojson/) map of the zones where you can also click on a zone to download the geojson, and  maybe put all the sites on there as well. 

not sure what else for 1.0

lboeman:
closes #279 
Adds climate zones support to the dashboard with a few features:
- Add climate zones to site listing in place of latitude and longitude
Screenshot from 2020-06-17 13-59-48

- Display the climate zones with the metadata on each site page

Screenshot from 2020-06-17 14-00-42

- Adds `sfa_dash.api_interface.climate_zones.py` for querying the api for climate zone data.


alorenzo175:
By default, I wouldn't include the geojson until it's needed. Save the downloading of a megabyte or two

lboeman:
> By default, I wouldn't include the geojson until it's needed. Save the downloading of a megabyte or two

I had thought that the top of the sites listing page was where the leaflet would live, thus loading the data there. Were you envisioning a separate climate zones page?

wholmgren:
@lboeman the interface changes look good to me. Did you want me to look at the code too?

Fine with me to make a follow up issue for a leaflet map.

lboeman:
> @lboeman the interface changes look good to me. Did you want me to look at the code too?
> 
> Fine with me to make a follow up issue for a leaflet map.

Interface review is great, thanks!. Code changes here are minimal,  most of the line count was merging templates that no longer made sense to exist separately. I'll open another Issue for the leaflet map and remove the geojson loading.

alorenzo175:
> > By default, I wouldn't include the geojson until it's needed. Save the downloading of a megabyte or two
> 
> I had thought that the top of the sites listing page was where the leaflet would live, thus loading the data there. Were you envisioning a separate climate zones page?

I'm not sure actually. More that I don't think we need to be grabbing that data and slowing the listing until we decide where and how to do a map, if at all

lboeman:
I've removed the climate zone loading here, and opened a PR for possibly adding the leaflet map in the future. Good to merge?

alorenzo175:
> I've removed the climate zone loading here, and opened a PR for possibly adding the leaflet map in the future. Good to merge?

good with me

PR #296, opened 2020-06-29
Add event to interval_label options

### LINKED ISSUE #295: cannot create event forecast
Screen Shot 2020-06-29 at 8 52 34 AM

lboeman:
closes #295 
Adds 'event' To interval label options and centralizes declaration of these options to `template_globals.py`

### DIFF
--- a/sfa_dash/templates/forms/cdf_forecast_group_form.html
+++ b/sfa_dash/templates/forms/cdf_forecast_group_form.html
@@ -44,9 +44,7 @@
 </div>
 <div class="form-element">
   {{ form.select('Interval label', 'interval_label',
-                 {'beginning': 'Beginning',
-                  'ending': 'Ending',
-                  'instant': 'Instant'},
+                 interval_label_options,
                   'Indicates if a time labels the start or the end of an interval.',
                   form_data=form_data) }}
 </div>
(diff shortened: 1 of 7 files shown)


sfa_dash/templates/forms/cdf_forecast_group_form.html
```diff
@@ -44,9 +44,7 @@
 </div>
 <div class="form-element">
   {{ form.select('Interval label', 'interval_label',
-                 {'beginning': 'Beginning',
-                  'ending': 'Ending',
-                  'instant': 'Instant'},
+                 interval_label_options,
                   'Indicates if a time labels the start or the end of an interval.',
```
wholmgren, line 48: this text should be changed to be more generic or to include instantaneous and event.

wholmgren: and be set for all of the forms

lboeman:
I updated the help text for interval label and added a little bit of JS to do the following:
- Select the `event` interval label (and hide others) if variable is set to `event`. Hide the `event` interval label otherwise.
- If the user selects a different variable, restore the previous selection they had made in interval_label to avoid errors when users don't realize the interval label has been reset.

PR #297, opened 2020-06-30
Add costs to reports

### LINKED ISSUE #289: Add cost to report creation
Not in core or API yet, but https://github.com/SolarArbiter/solarforecastarbiter-core/blob/7626b9ed1e46d6552df00dbf8cffab0fe2b69dba/solarforecastarbiter/tests/data/report_metadata.json can be used as a jumping off point for an idea of what the JSON from teh API will need to look like

lboeman:
closes #289 

Still a work in progress, but mostly functioning time of day, constant and datetime costs.
Needs:

- [x] Add support for error band costs. I think the biggest thing here will be keeping the dom and user interaction reasonable. @alorenzo, when specifying error band costs are the bands expected to be contiguous over (-Inf, Inf) or can there be gaps? They should not overlap, correct? I wasn't sure if the first suggested error range should be (-Inf, Inf) and then, when the user sets a min or max the next band they add will start with (-Inf, max) or (min, Inf).
- [x] Add api -> form data conversion.
- [x] Touch up the style a little bit.

Screenshots of current state below. 
tod_params
dt_params
constant_params

### DIFF
--- a/sfa_dash/form_utils/converters.py
+++ b/sfa_dash/form_utils/converters.py
@@ -443,14 +443,16 @@ class ReportConverter(FormConverter):
         fill = form_data[f'cost-fill{suffix}']
         net = form_data.get(f'cost-net{suffix}', False)
         timezone = form_data[f'cost-timezone{suffix}']
-        return {
+        payload = {
             'datetimes': datetimes,
             'cost': costs,
             'aggregation': aggregation,
             'fill': fill,
             'net': net,
-            'timezone': timezone,
         }
+        if timezone is not None:
+            payload.update({'timezone': timezone})
+        return payload
 
     @classmethod
     def parse_form_tod_cost(cls, form_data, index=None):
@@ -464,14 +466,17 @@ class ReportConverter(FormConverter):
         fill = form_data[f'cost-fill{suffix}']
         net = form_data.get(f'cost-net{suffix}', False)
         timezone = form_data[f'cost-timezone{suffix}']
-        return {
+        payload = {
             'times': times,
             'cost': costs,
             'aggregation': aggregation,
             'fill': fill,
             'net': net,
-            'timezone': timezone,
         }
+        if timezone is not None:
+            payload.update({'timezone': timezone})
+        return payload
+
 
     @classmethod
     def parse_form_errorband_cost(cls, form_data, index=None):
@@ -512,12 +517,12 @@ class ReportConverter(FormConverter):
     def parse_form_costs(cls, form_data):
         """Parses costs from form data into an api-conforming dictionary.
         """
-        cost_name = form_data.get('master-cost-name')
+        cost_name = form_data.get('cost-primary-name')
 
         # Check for existence of cost name before trying to parse the rest,
         # as cost is an optional attribute.
         if cost_name is not None:
-            cost_type = form_data['master-cost-type']
+            cost_type = form_data['cost-primary-type']
             parameter_parser = cls.get_form_cost_parser(cost_type)
             cost_parameters = parameter_parser(form_data)
             return [{
--- a/sfa_dash/form_utils/converters.py
+++ b/sfa_dash/form_utils/converters.py
@@ -403,11 +403,11 @@ class ReportConverter(FormConverter):
     @classmethod
     def extract_tods(cls, value):
         """Parses time of day values from a csv string"""
-        return value.split(',')
+        return [v.strip() for v in value.split(',')]
 
     @classmethod
     def extract_datetimes(cls, value):
-        return value.split(',')
+        return [v.strip() for v in value.split(',')]
 
     @classmethod
     def extract_many_costs(cls, value):
@@ -450,7 +450,7 @@ class ReportConverter(FormConverter):
             'fill': fill,
             'net': net,
         }
-        if timezone is not None:
+        if timezone is not None and timezone != 'null':
             payload.update({'timezone': timezone})
         return payload
 
@@ -473,10 +473,9 @@ class ReportConverter(FormConverter):
             'fill': fill,
             'net': net,
         }
-        if timezone is not None:
+        if timezone is not None and timezone != 'null':
             payload.update({'timezone': timezone})
         return payload
-
 
     @classmethod
     def parse_form_errorband_cost(cls, form_data, index=None):
@@ -498,7 +497,7 @@ class ReportConverter(FormConverter):
                 'cost_function': cost_function,
                 'cost_function_parameters': parameters,
             })
-        return bands
+        return {'bands': bands}
 
     @classmethod
     def get_form_cost_parser(cls, cost_type):
--- a/sfa_dash/form_utils/converters.py
+++ b/sfa_dash/form_utils/converters.py
@@ -362,7 +362,7 @@ class AggregateConverter(FormConverter):
 
 class ReportConverter(FormConverter):
     @classmethod
-    def zip_object_pairs(cls, form_data, cost_name=None):
+    def zip_object_pairs(cls, form_data):
         """Create a list of object pair dictionaries containing a
         forecast and either an observation or aggregate.
         """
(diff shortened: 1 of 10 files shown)


alorenzo175:
No, the ranges can overlap, and the order matters. Maybe https://solarforecastarbiter-core.readthedocs.io/en/latest/cost.html is clear enough?

lboeman:
Yeah, that helps a lot. Thanks!

On Jun 29, 2020 7:40 PM, Tony Lorenzo <notifications@github.com> wrote:

External Email

No, the ranges can overlap, and the order matters. Maybe https://solarforecastarbiter-core.readthedocs.io/en/latest/cost.html is clear enough?

lboeman:
@alorenzo175 @wholmgren This is getting close. I have a question regarding Datetime costs and timezone. Currently, I've relaxed the datetime validation for iso8601 to allow for omitting dashes in dates and colons in times. I still have two hard requirement: the trailing 'Z' to indicate UTC, and that the time part of the datetime only contains `HH:MM` or `HHMM`. Should I omit the Z for datetime costs and set the timezone to the cost's timezone when parsing?

alorenzo175:
> Should I omit the Z for datetime costs and set the timezone to the cost's timezone when parsing?

I'm not sure exactly what you mean. In core, if the datetimes are already localized, that tz is used, otherwise the cost timezone is used, otherwise the observation timezone is used. So we should allow the  datetimes to be unlocalized (no Z) or localized (Z or -05:00 etc)

lboeman:
Alright, after the last comment I realized that sticking with the super strict datetime validator was silly , so I've updated it to allow for Z, an offset or unlocalized time as Tony suggested. This won't catch correctly formatted invalid dates, but those will be caught by the API. 
I've added Error Band support with a screenshot below. The interface is simple: click "Add Error Band" and it adds a new set of fields where you can change the band type and set an error range. Currently the only validation on the error range is start < end and value is a float (including +/- Infinity).

errorband

alorenzo175:
Looks good at first look. Minor style things that might be quick: the checkboxes are too tight to the text for me and only time of day is not bold for error band

alorenzo175:
And how hard is it to add simple help popouts for the parameters like net and aggregate?

lboeman:
Thanks for checking it out. Help text shouldn't be too bad. I'll touch that up along with your style points.

alorenzo175:
A few more things: 
- timezone should be an option for both time of  day and datetime
- a none/null timezone should be the default (although you might not be able to post that to the api)
- I guess all datetimes do need to have the same timezone. Maybe easiest is to require Z/UTC and fix this later https://sentry.io/organizations/solararbiter-rx/issues/1755701915/?project=5254377&referrer=alert_email
- 

Things that can probably wait:
- on a post error, the cost fields disappear until you select a different cost type so whatever parameters you entered (e.g. times) are lost

lboeman:
> * somehow cost made it into metrics twice in https://dev-dashboard.solarforecastarbiter.org/reports/cbe1dc8a-bb22-11ea-9309-0a580a8002e6

Ahh, I added code to append it if a cost is defined because when I started it wasn't in core's `ALLOWED_DETERMINISTIC_METRICS`, woops. On that note, Is it a valid metric for events/probabilistic forecasts?

wholmgren:
Is white space allowed in the csv fields?

Can the bands be reordered once created? Maybe something for future work.

alorenzo175:
> > * somehow cost made it into metrics twice in https://dev-dashboard.solarforecastarbiter.org/reports/cbe1dc8a-bb22-11ea-9309-0a580a8002e6
> 
> Ahh, I added code to append it if a cost is defined because when I started it wasn't in core's `ALLOWED_DETERMINISTIC_METRICS`, woops. On that note, Is it a valid metric for events/probabilistic forecasts?


Nope, only deterministic right now, so the parameters should show up on those report forms

sfa_dash/form_utils/converters.py
```diff
@@ -400,6 +400,131 @@ def parse_api_filters(cls, report_parameters):
...
+
+    @classmethod
+    def extract_datetimes(cls, value):
+        return value.split(',')
+
+    @classmethod
+    def extract_many_costs(cls, value):
+        costs = value.split(',')
+        costs = [float(cost) for cost in costs]
+        return costs
```
alorenzo175, line 416: for all of these `extract_`, when empty '' is passed so at least `float` raises an error

sfa_dash/form_utils/converters.py
```diff
@@ -436,6 +564,13 @@ def formdata_to_payload(cls, form_dict):
         report_params['end'] = form_dict['period-end']
         report_params = cls.apply_crps(report_params)
         report_dict = {'report_parameters': report_params}
+        if len(costs) > 0:
+            cost_name = costs[0]['name']
+            report_params['metrics'].append('cost')
```
alorenzo175, line 569: I guess check if added first. Although it should probably get checked automatically on the dashboard so it's not opaque to the user

sfa_dash/templates/forms/base/base_report_form.html
```diff
@@ -42,6 +42,17 @@ <h3>Create New Report</h3>
            The end of the analysis period as an ISO 8601 datetime.
        </span>
      </div>
+         <h5>Cost Parameters</h5>
```
alorenzo175, line 45: a bit more space above this

sfa_dash/form_utils/converters.py
lboeman: > Is white space allowed in the csv fields?
> 
> Can the bands be reordered once created? Maybe something for future work.

Whitespace should be allowed, but I'll test it out.
Bands can't be reordered right now, but I agree valuable feature for the future. I'm not sure how simple it would be to implement.

lboeman:
> adding a time of day error band, then clicking add error band, then clicking on the datetime text (not circle) selects datetime as the first band and info is lost

This was from adding the `for` attribute to labels. I've removed those attributes.

lboeman:
I've got to all of the points, but I found a new issue. In it's current state, can't submit a report without cost. yay.

wholmgren:
> Bands can't be reordered right now, but I agree valuable feature for the future. I'm not sure how simple it would be to implement.

I was thinking that you'd add up and down arrows next to remove, and clicking them would move the `<div>` container one spot.

lboeman:
I've updated this such that cost parameters are only displayed when the user selects the `cost` metric. There are two reasons for doing this: 

1. It's confusing to the user to define costs, and then have to select another button to apply them. This makes it such that if you select cost as a metric, it displays the required fields.

2. All cost fields are required if *any* are supplied. This means that we need to make sure that those fields are not required if the user decides against adding cost. This is really difficult to manage if they got halfway through a cost form and decided to abandon it, you have to deal with invalid states they left in fields etc.

Any selections a user has made should be persisted between when they select/deselect the cost metric. The form is backed by some js classes that exist in the global "cost" variable, and should update whenever the user changes the form.

Also fixed a few bugs with prefilling the form, so it should work as expected on error/cloning. 

What do you guys think?

lboeman:
Alright, think I've hit all the notes.
- Added help to timezone, name and cost functions (added label for cost function radio buttons).
- Cost parameters header now has the same style as other section headers e.g. 'Quality Flags'.
- Wrapped cost parameters in a field set and added a small arrow for expanding/collapsing. A user may expand the cost parameters prior to selecting cost, which displays the disabled form.
disabled
When the user checks the cost box, the parameters will open automagically and if there is an existing cost object, the fields will be filled. 
enabled

wholmgren:
I skimmed the code and it looked ok but I easily could have missed something. That's a lot of javascript.

alorenzo175:
I have not looked at the javascript. But in the console, looks like there is a log statement somewhere in there. And not sure why, but the validation errors do not show up in my browser. The field turns red but that's it. Firefox 77.0.1

alorenzo175:
I guess the validation errors also don't show up on from the current dev dashboard. Just noticed it now since I saw them on your screen, so maybe a separate issue.

lboeman:
> I have not looked at the javascript. But in the console, looks like there is a log statement somewhere in there. And not sure why, but the validation errors do not show up in my browser. The field turns red but that's it. Firefox 77.0.1

This is a frustrating issue, I can replicate this in firefox as well. Calling `element.reportValidity()` from the console makes it pop up, but it doesn't work from the code.

wholmgren:
Not a lot of firefox users, ok to make it a follow up issue.

lboeman:
Not sure why the issue with manually calling `reportValidity()`, but the messages do seem to at least pop up when you try to submit the form.

PR #302, opened 2020-07-01
replace special characters that are in prob. fx metrics

### LINKED ISSUE #300: Report javascript error for probabilistic forecasts
image

alorenzo175:
along with https://github.com/SolarArbiter/solarforecastarbiter-core/pull/507 closes #300


lboeman:
Does this run for you? I get `Uncaught TypeError: value.replace(...).replaceAll is not a function`. when i run this locally on Chrome.

lboeman:
Ahh yeah, this works in Firefox and chrome after an update. Support seems spotty though. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#Browser_compatibility
probably need to stick with using `.replace(/x/g, 'thing')` for now.

alorenzo175:
ah that sucks

PR #308, opened 2020-07-01
Misc style stuff

### LINKED ISSUE #305: layout issue for account drop down
Screen Shot 2020-07-01 at 1 55 42 PM

my window is 1270x1338px. firefox 77.0.1

lboeman:
Closes a few minor style issues.
820ae65 closes #305

### DIFF
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -279,7 +279,7 @@ class RecomputeReportView(BaseView):
         except DataRequestException as e:
             self.flash_api_errors(e.errors)
         else:
-            flash('Report recomputed successfully.', 'message')
+            flash('Scheduled report recompute successfully.', 'message')
         return redirect(url_for('data_dashboard.report_view',
                                 uuid=uuid))
 
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -279,7 +279,9 @@ class RecomputeReportView(BaseView):
         except DataRequestException as e:
             self.flash_api_errors(e.errors)
         else:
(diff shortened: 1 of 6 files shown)


lboeman:
Here are the proposed fixes for the issues closed in this PR:
- Updated report recompute message to say "Scheduled report recompute successfuly."
- Page headers all use the same `<h2>` tag.
- Right justified the Account dropdown, I had to enforce a fixed width to get this to work so there may be wrapping for some users with really long emails.
justified_account_menu
- Added some space between select all/none buttons in column filters.
filter_spacing

sfa_dash/blueprints/reports.py
```diff
@@ -279,7 +279,7 @@ def get(self, uuid):
         except DataRequestException as e:
             self.flash_api_errors(e.errors)
         else:
-            flash('Report recomputed successfully.', 'message')
+            flash('Scheduled report recompute successfully.', 'message')
```
wholmgren, line 282: Should we add text along the lines of wait 30 seconds and refresh the page? Or add a timer? (like we have for the original waiting page)

lboeman: The user should still land back on that page and be able to see that message. I believe there is also a script Tony added to automagically reload the page on a timer.

wholmgren: I'm confused. This just adds a box to the previously rendered report. It doesn't send the user to another page.

lboeman: Ah, I was thinking that the report status was reset to pending on recompute. Now I see that it continues to serve the old report until it gets recomputed. Updated the text.

PR #311, opened 2020-07-01
Forecast fill

### LINKED ISSUE #292: Add controls for Forecast filling
In support of https://github.com/SolarArbiter/solarforecastarbiter-core/pull/493 
Will need a way for users to define how to fill missing forecasts.

lboeman:
Adds a forecast fill field at the end of all of the report forms. UI is simple radio buttons as suggested by @wholmgren.
closes #292 

When `drop` or `forward` options  are selected, the field for providing a value is disabled.
Screenshot from 2020-07-01 16-54-07

When the 'filled with' option is selected the field is activated and required. There is a bit of js that sets the min max and step to 0, 1, and 1 respectively for events.
Screenshot from 2020-07-01 16-53-59

### DIFF
--- a/sfa_dash/form_utils/converters.py
+++ b/sfa_dash/form_utils/converters.py
@@ -597,7 +597,7 @@ class ReportConverter(FormConverter):
         form_params['metrics'] = report_parameters.get('metrics', [])
         form_params['period-start'] = report_parameters['start']
         form_params['period-end'] = report_parameters['end']
-        form_params['costs'] = report_parameters.get('costs', [])
+        form_params['costs'] = report_parameters['costs']
         form_params['forecast_fill_method'] = report_parameters.get(
             'forecast_fill_method',
             'drop'
--- a/sfa_dash/form_utils/converters.py
+++ b/sfa_dash/form_utils/converters.py
@@ -557,6 +557,13 @@ class ReportConverter(FormConverter):
             return params
 
     @classmethod
+    def parse_fill_method(cls, form_dict):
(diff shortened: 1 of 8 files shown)


sfa_dash/static/js/report-form-utilities.js
```diff
@@ -1421,3 +1421,20 @@ report_utils.help_popup = function(help_name, help_text){
     var help_box = $(`<span class="${help_name}-help-text form-text text-muted help-text collapse" aria-hidden="">${help_text}</span>`);
     return [help_button, help_box]
 }
+report_utils.register_forecast_fill_method_validator = function(
+    forecast_type='deterministic'){
```
wholmgren, line 1425: maybe make this `''` to avoid potential confusion on the probabilistic form. Or pass the type in all of the register calls.

sfa_dash/form_utils/converters.py
```diff
@@ -588,7 +597,11 @@ def payload_to_formdata(cls, payload_dict):
         form_params['metrics'] = report_parameters.get('metrics', [])
         form_params['period-start'] = report_parameters['start']
         form_params['period-end'] = report_parameters['end']
-        form_params['costs'] = report_parameters['costs']
+        form_params['costs'] = report_parameters.get('costs', [])
```
wholmgren, line 600: looks unrelated to this PR. that's ok, just double checking it's not a merge issue.

alorenzo175:
Nope

PR #315, opened 2020-07-06
fire forecast filtering when units are set

### LINKED ISSUE #314: report form does not exclude other variables after selecting first forecast pair
After selecting the first forecast/observation pair, the dashboard form sets the forecast variable but the menu is not updated:

Screen Shot 2020-07-06 at 9 25 40 AM

Manually adjusting the menu does work:

Screen Shot 2020-07-06 at 9 25 52 AM

lboeman:
Closes #314 
Adds a callback argument to the `report_utils.set_units` and `report_utils.unset_units` that will be called after setting the units/variables. The forecast filtering functions are passed in when setting/unsetting the units.

### DIFF
--- a/sfa_dash/static/js/report-form-utilities.js
+++ b/sfa_dash/static/js/report-form-utilities.js
@@ -152,14 +152,17 @@ report_utils.toggle_reference_dependent_metrics = function(){
 current_units = null;
 
 
-report_utils.unset_units = function(){
+report_utils.unset_units = function(filter_callback=null){
     current_units = null;
     // Call to setVariables to repopulate the options in the Variable <select>
     // element.
     report_utils.setVariables()
+    if (filter_callback){
+        filter_callback();
+    }
 }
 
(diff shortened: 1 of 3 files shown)


sfa_dash/static/js/report-form-utilities.js
```diff
@@ -152,14 +152,17 @@ report_utils.toggle_reference_dependent_metrics = function(){
 current_units = null;
 
 
-report_utils.unset_units = function(){
+report_utils.unset_units = function(filter_callback=null){
```
wholmgren, line 155: what's the purpose of `=null` here and below?

lboeman: To make the filtering optional argument. Right now, it's always called with a callback. Due to recently kicking around the idea of merging all three types of report forms I thought this should be an optional argument because I don't know how that'll all tie together.

PR #317, opened 2020-07-06
trigger filter forecast through site change callback to avoid scoping…

### LINKED ISSUE #316: unit restriction not unset when clearing all obs/forecast pairs
Screen Shot 2020-07-06 at 10 55 02 AM

follow up to #314 #315

lboeman:
issue
closes #316


lboeman:
This activates the forecast filtering after unset with the site selector's onchange callback due to the `filterForecast` function being defined as a closure inside the larger init function. The onchange callback just calls the filterForecast function and works whether sites or observations are being compared to.

PR #319, opened 2020-07-06
Fix climate zone filtering

### LINKED ISSUE #318: Sites with no climate zone break table filters
Because an empty value matches everything, deselecting the empty climate zone breaks the filter. Should create a meaningful label for display, and a value to check against.

### LINKED ISSUE #321: add link to climate zones page
I have no idea what the reference regions are without looking at them at https://solarforecastarbiter.org/climatezones/ 

Two ideas for how to improve this. 

## Idea 1

In the header row add a link to climate zones description page like this:

Screen Shot 2020-07-06 at 5 59 53 PM

## Idea 2

Turn the text for each climate zone into a link to the relevant zone. For example: [Reference Region 3](https://solarforecastarbiter.org/climatezones/#region3)

lboeman:
closes #318 
closes #321 
Inserts "No Climate Zone" when a site's climate_zones attribute is an empty list, so that the value can be filtered on. This also sorts all of the filter options for *every* table filter.
Screenshot from 2020-07-06 16-30-13

### DIFF
--- a/sfa_dash/filters.py
+++ b/sfa_dash/filters.py
@@ -84,10 +84,10 @@ def climate_zone_links(zones):
             return f'<a href="https://solarforecastarbiter.org/climatezones/#region{zone_number}">{zone_name}</a>'  # noqa: E501
         else:
             return zone_name
-    if not zones:
-        return ["No Climate Zone"]
-    else:
+    if zones:
         return [reference_zone_link(zone) for zone in zones]
+    else:
+        return ["No Climate Zone"]
 
 
 def register_jinja_filters(app):
--- a/sfa_dash/filters.py
+++ b/sfa_dash/filters.py
@@ -85,9 +85,9 @@ def climate_zone_links(zones):
         else:
             return zone_name
     if not zones:
-        return "No Climate Zone"
+        return ["No Climate Zone"]
     else:
-        return ', '.join([reference_zone_link(zone) for zone in zones])
+        return [reference_zone_link(zone) for zone in zones]
 
 
 def register_jinja_filters(app):
--- a/sfa_dash/filters.py
+++ b/sfa_dash/filters.py
@@ -77,6 +77,19 @@ def is_number(num):
     return str(num).isnumeric()
 
 
+def climate_zone_links(zones):
+    def reference_zone_link(zone_name):
+        if 'Reference Region' in zone_name:
+            zone_number = zone_name.split(' ')[-1]
+            return f'<a href="https://solarforecastarbiter.org/climatezones/#region{zone_number}">{zone_name}</a>'  # noqa: E501
+        else:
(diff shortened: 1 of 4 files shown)


sfa_dash/templates/data/table/site_table.html
```diff
@@ -20,7 +20,12 @@
     <tr>
       <td class="name-column"><a href="{{ row['link'] }}">{{ row['name'] }}</a></td>
       <td class="provider-column">{{ row['provider'] }}</td>
-      <td class="climate-zones-column">{{ row['climate_zones'] | join(', ') }}</td>
+      {% if row['climate_zones'] |length > 0 %}
+      {% set climate_zones = row['climate_zones'] |join(', ') %}
```
wholmgren, line 24:
```suggestion
      {% if row['climate_zones'] | length > 0 %}
      {% set climate_zones = row['climate_zones'] | join(', ') %}
```

sfa_dash/templates/data/table/site_table.html
```diff
@@ -20,7 +20,12 @@
...
+      {% else %}
+      {% set climate_zones = "No Climate Zone" %}
```
wholmgren, line 26: Should we define this at a higher level so that it can be referenced on, say, a site page? I don't know where to suggest. I also wouldn't list this like a normal zone on a site page, so maybe just ignore me and leave as is for now.

lboeman:
closes #321 
Added links to the climate zones in the site table with a new template filter that contains the if/else logic that was previously in the template. This only adds links for zones containing "Reference Region" so that if we start adding custom zones it should still work.

lboeman:
Also linked the climate zones on site pages.

sfa_dash/templates/data/table/site_table.html
lboeman: Updates to this move the linking logic to a filter so that it can also be used on site pages, but only print the zones if the list isn't empty, so "No Climate Zones" is not displayed.

sfa_dash/filters.py
```diff
@@ -77,10 +77,24 @@ def is_number(num):
...
 
 
+def climate_zone_links(zones):
+    def reference_zone_link(zone_name):
+        if 'Reference Region' in zone_name:
+            zone_number = zone_name.split(' ')[-1]
+            return f'<a href="https://solarforecastarbiter.org/climatezones/#region{zone_number}">{zone_name}</a>'  # noqa: E501
+        else:
+            return zone_name
+    if not zones:
```
wholmgren, line 87: If we can assume that `zones` is always a python list instead of numpy array then reverse the logic so this reads `if zones: ... else: return ["No Climate Zone"]`. If we cannot assume that, then the statement should be `if zones is not None:`

PR #320, opened 2020-07-07
Retain variable selection through object pair selection

lboeman:
Variable selector was losing selected value and reverting to the first visible value in the list whenever a pair was added. Not sure exactly why this behavior was happening, but this code ensures that if a value is still visible, it is retained.

### DIFF
--- a/sfa_dash/static/js/report-form-utilities.js
+++ b/sfa_dash/static/js/report-form-utilities.js
@@ -207,7 +207,6 @@ report_utils.setVariables = function(){
      * current units.
      */
     var selected_var = $('#variable-select').val();
-    console.log(selected_var);
     var variable_options = $('#variable-select option');
     variable_options.removeAttr('hidden');
     if (current_units){
--- a/sfa_dash/static/js/report-form-utilities.js
+++ b/sfa_dash/static/js/report-form-utilities.js
@@ -203,24 +203,30 @@ report_utils.searchObjects = function(object_type, object_id){
 }
 
 report_utils.setVariables = function(){
-	/* Displays or hides options in the variable <select> element based on the
-     * current units.
+	/*  Displays or hides options in the variable <select> element based on the
+     *  current units.
      */
     var selected_var = $('#variable-select').val();
     var variable_options = $('#variable-select option');
     variable_options.removeAttr('hidden');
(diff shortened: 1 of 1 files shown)


sfa_dash/static/js/report-form-utilities.js
```diff
@@ -206,19 +206,23 @@ report_utils.setVariables = function(){
 	/* Displays or hides options in the variable <select> element based on the
      * current units.
      */
-    variable_options = $('#variable-select option');
+    var selected_var = $('#variable-select').val();
+    console.log(selected_var);
```
wholmgren, line 210: debug code?

lboeman:
> I'd need to spend more time engaging with the javascript to understand this. Does it work and is it thoroughly tested?

I poked it some more and tested to confirm the following behavior:

- When a pair is added, the variable options are limited to variables with the same units and the variable of the pair is selected from the variable drop down menu. The behavior before was that the first variable with matching units was selected.
- When the last pair is removed, the variable drop down reverts to "All Variables" and retains the forecast selection.

sfa_dash/static/js/report-form-utilities.js
```diff
@@ -202,23 +202,35 @@ report_utils.searchObjects = function(object_type, object_id){
...
 
-report_utils.setVariables = function(){
-	/* Displays or hides options in the variable <select> element based on the
-     * current units.
+report_utils.setVariables = function(variable=null){
+	/*  Displays or hides options in the variable <select> element based on the
+     *  current units.
+     *
+     *  @param {string} variable
+     *      The variable to remain selected after filtering. Ignored when
```
wholmgren, line 210: The variable to remain? Multiple variables remain if they share units.

lboeman: Yes, but for example after adding a pair with variable 'ghi', this ensures that 'ghi' is what remains in the variables filter after filtering takes place. This does change the select from "All Variables" to that value, which I can see is somewhat confusing.

We could allow "All Variables" to remain an option while filtering the forecasts by the current units but that would take some more work. The way filtering works now, forecasts are filtered by the currently selected variable, and the options for the variable filter are determined by units. Hopefully that makes sense.

PR #327, opened 2020-07-14
Fix permissions with object type 'report' 

### LINKED ISSUE #326: Permissions for report objects TypeError
Sentry Issue: [DASHBOARD-34](https://sentry.io/organizations/solararbiter-rx/issues/1784291339/?referrer=github_integration)

```
TypeError: 'NoneType' object is not subscriptable
(5 additional frame(s) were not displayed)
...
  File "flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "flask/views.py", line 88, in view
    return self.dispatch_request(*args, **kwargs)
  File "flask/views.py", line 158, in dispatch_request
    return meth(*args, **kwargs)
  File "sfa_dash/blueprints/admin.py", line 667, in get
    all_objects = [obj for obj in all_objects
  File "sfa_dash/blueprints/admin.py", line 668, in <listcomp>
    if obj[object_id_key] not in perm_objects]
```

lboeman:
closes #326 
Also fixes an issue that occurred with unsafe permission descriptions being printed. Discovered that inside macros, global autoescape rules are not respected, so added explicit autoescape blocks to macros that print user provided data. We should also probably add the same user string validation to the description field at the API level.

### DIFF
--- a/sfa_dash/blueprints/admin.py
+++ b/sfa_dash/blueprints/admin.py
@@ -655,12 +655,11 @@ class PermissionObjectAddition(PermissionView):
             api = self.get_api_handler(permission['object_type'])
             perm_objects = list(permission['objects'].keys())
 
-            if data_type == 'report':
-                all_objects = api.list_metadata()
-                all_objects = self.flatten_reports(all_objects)
-            else:
-                all_objects = api.list_metadata()
+            all_objects = api.list_metadata()
             all_objects = self.filter_by_org(all_objects, 'provider')
+
+            if data_type == 'report':
+                all_objects = self.flatten_reports(all_objects)
             # remove any objects alread on the permission
             object_id_key = f"{data_type}_id"
             all_objects = [obj for obj in all_objects
--- a/sfa_dash/blueprints/admin.py
+++ b/sfa_dash/blueprints/admin.py
@@ -66,9 +66,9 @@ class AdminView(BaseView):
         self.template_args = {'subnav': self.format_subnav(**subnav_kwargs)}
 
     def flatten_reports(self, report_list):
-        """Takes a list of report objects from the api and updates the
-        top-level keys with the contents of 'report_parameters' for each
-        report.
+        """Takes a list of report metadata dictionaries and moved the `name`
+        field to the top-level dictionary and removes unneeded report
+        parameters and raw report.
         """
         for report in report_list:
             report.update({'name': report.pop('report_parameters')['name']})
--- a/sfa_dash/blueprints/admin.py
+++ b/sfa_dash/blueprints/admin.py
@@ -72,6 +72,7 @@ class AdminView(BaseView):
         """
         for report in report_list:
             report.update({'name': report.pop('report_parameters')['name']})
+            report.pop('raw_report', None)
         return report_list
 
     def get(self):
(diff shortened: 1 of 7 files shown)


lboeman:
Also fixed the issue where a Permissions with `object_type` of report did not display the associated reports. The issue was that the list of reports was not being properly formatted into a `{uuid: metadata, ...}`, this has been corrected and now reports will be populated as with other object types.

sfa_dash/templates/forms/form_macros.jinja
```diff
@@ -17,7 +17,15 @@
...
   <label>{{ label }}</label><br/>
   <div class="input-wrapper">
-    <input type="{{ type }}" class="form-control {{ class }} {{ name }}-field" {{ extra_attr }}  name="{{ name }}"  {% if name == 'name' %}{{ name_validation() }}{% endif %} {% if form_data is not none %}value="{{ form_data.get(name, default) }}"{% endif %}>{% if units is not none %}{{ units }}{% endif %}
+    <input type="{{ type }}"
+     class="form-control {{ class }} {{ name }}-field"
+     {{ extra_attr }}
+     name="{{ name }}"
+     {% if name == 'name' %}{{ name_validation() }}{% endif %}
+     {% if form_data is not none %}
+     value="{% autoescape true %}{{ form_data.get(name, default) }}{% endautoescape %}"
```
alorenzo175, line 26: Can we just escape the variable? https://jinja.palletsprojects.com/en/2.11.x/templates/#working-with-manual-escaping

sfa_dash/blueprints/admin.py
```diff
@@ -655,20 +654,21 @@ def get(self, uuid, **kwargs):
...
             api = self.get_api_handler(permission['object_type'])
             perm_objects = list(permission['objects'].keys())
+
             if data_type == 'report':
                 all_objects = api.list_metadata()
-                all_objects = [rep.update(rep['report_parameters'])
-                               for rep in all_objects]
+                all_objects = self.flatten_reports(all_objects)
             else:
                 all_objects = api.list_metadata()
```
alorenzo175, line 662: move this out of the if statement

alorenzo175: and then we can also filter before flattening reports, yes?

sfa_dash/blueprints/admin.py
```diff
@@ -72,7 +72,7 @@ def flatten_reports(self, report_list):
         """
         reports = [report.copy() for report in report_list]
         for report in reports:
-            report.update(report.get('report_parameters'))
+            report.update(report.pop('report_parameters'))
```
alorenzo175, line 75: do we need the raw_report for anything here? good to drop if we don't

alorenzo175: Actually, now I don't understand the purpose of this function for the admin pages. What parameters of a report are necessary besides the name, provider, and id?

lboeman: It's really just moving all of the keys in report parameters to the top level of the report metadata dictionaries. You're right that most aren't necessary, this could be reduced to just move the name field.

PR #339, opened 2020-07-16
Catch role sharing error

### LINKED ISSUE #335: crash when role grant fails
Sentry Issue: [DASHBOARD-36](https://sentry.io/organizations/solararbiter-rx/issues/1786878730/?referrer=github_integration)

```
DataRequestException: 
  File "sfa_dash/blueprints/admin.py", line 310, in post
    users.add_role_by_email(user_email, uuid)
  File "sfa_dash/api_interface/users.py", line 36, in add_role_by_email
    payload=None)
  File "sfa_dash/api_interface/__init__.py", line 69, in post_request
    f'{app.config["SFA_API_URL"]}{path}', **kwargs))
  File "sfa_dash/api_interface/util.py", line 57, in handle_response
    raise DataRequestException(request_object.status_code, **errors)

TypeError: set_template_args() missing 2 required positional arguments: 'role' and 'redirect_link'
(4 additional frame(s) were not displayed)
...
  File "flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "flask/views.py", line 88, in view
    return self.dispatch_request(*args, **kwargs)
  File "flask/views.py", line 158, in dispatch_request
    return meth(*args, **kwargs)
  File "sfa_dash/blueprints/admin.py", line 320, in post
    self.set_template_args()
```

### LINKED ISSUE #336: Site permission creation fails
Failure occurs when trying to create a new site permission, user remains at the correct path. `/admin/permissions/create/site` but observation options are displayed.
Screenshot from 2020-07-16 10-51-00

lboeman:
Left out arguments to `set_template_args()` when attempt to share a role fails. I went through the admin module to try to catch any other issues, and added comments where I thought they were needed. 

closes #335 

closes #336 - Issue was missing closing quotation mark when printing the checkbox values in the site_row macro. This was causing a trailing space to be submitted with the site uuid, causing a crash wehn trying to add the site to a permission. 

Also altered the behavior if a failure occurs while adding objects to a permission during permission creation. Instead of leaving users on the creation form, they are directed to the permission page, where they can see it's metadata and the objects that were successfully added, with error messages about the objects that failed. Note that if the creation of the permission fails, the user remains on the form and receives an error message. 

Also added the newly added name javascript validation to the role creation form.

### DIFF
--- a/sfa_dash/blueprints/admin.py
+++ b/sfa_dash/blueprints/admin.py
@@ -306,28 +306,27 @@ class RoleGrant(AdminView):
     def post(self, uuid):
         form_data = request.form
         user_email = form_data.get('user_email', '')
+        redirect_url = session.pop(
+            'redirect_link',
+            url_for('admin.role_view', uuid=uuid))
         try:
             users.add_role_by_email(user_email, uuid)
         except DataRequestException:
             # flash a message that grant failed
-            flash('Failed to grant role', 'error')
+            flash('Failed to grant role.', 'error')
             try:
                 role = roles.get_metadata(uuid)
             except DataRequestException as e:
                 # User could not read the role, flash a 404
                 self.flash_api_errors(e.errors)
                 role = None
-            self.set_template_args()
+            self.set_template_args(role, redirect_url)
             return render_template(
                 self.template,
-                role=role,
                 form_data=form_data,
                 **self.template_args)
         # flash success message and redirect to the referrer, or the role view
         flash('Role granted successfully', 'message')
-        redirect_url = session.pop(
-            'redirect_link',
-            url_for('admin.role_view', uuid=uuid))
         return redirect(redirect_url)
 
 
--- a/sfa_dash/blueprints/admin.py
+++ b/sfa_dash/blueprints/admin.py
@@ -174,10 +174,10 @@ class UserRoleRemoval(AdminView):
     def get(self, uuid, role_id, **kwargs):
         """Confirmation view for removing a role from a user
         """
-        redirect_link = request.headers.get(
-            'Referer', url_for('admin.roles'))
         # set a redirect link, because we can be directed here
         # from a role or user page.
+        redirect_link = request.headers.get(
+            'Referer', url_for('admin.roles'))
         try:
             user = users.get_metadata(uuid)
         except DataRequestException:
@@ -202,6 +202,8 @@ class UserRoleRemoval(AdminView):
             _external=True,
             uuid=uuid,
             role_id=role_id)
+        # If the user was not sent here from the confirmation page,
+        # redirect them.
         if request.headers.get('Referer') != confirmation_url:
             redirect(confirmation_url)
         try:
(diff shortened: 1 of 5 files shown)


sfa_dash/templates/forms/admin/permission_macros.jinja
```diff
@@ -20,7 +20,7 @@
 {% autoescape true %}
 {% set uuid = row_values['site_id'] %}
 <tr visible="true">
-  <td><input type="checkbox" name="objects-list-{{ index }}" value="{{ uuid }} {% if checked %}checked{% endif %}"></td>
+  <td><input type="checkbox" name="objects-list-{{ index }}" value="{{ uuid }}" {% if checked %}checked{% endif %}"></td>
```
wholmgren, line 23: stray " at end?

sfa_dash/templates/forms/admin/role_grant.html
```diff
@@ -10,12 +10,13 @@ <h3>Granting role "{{ role['name'] }}"</h3>
...
-    <input type="email" class="form-control" name="user_email" placeholder="Email">
+    {% if form_data is not defined %}{% set form_data = {} %}{% endif %}
+    <input type="email" class="form-control" name="user_email" placeholder="Email" value="{{ form_data.get('user_email', '')}}">
   </div>
   {{ form.token() }}
   </div>
 </form>
 <button type="submit" form="role-grant-form" value="Submit" class="btn btn-danger btn-sml">Grant Role</button>
-<a href="{{ redirect_link }}" class="btn btn-primary">Cancel</a>
+<a href="{{ redirect_link |e }}" class="btn btn-primary">Cancel</a>
```
wholmgren, line 20:
```suggestion
<a href="{{ redirect_link | e }}" class="btn btn-primary">Cancel</a>
```

sfa_dash/blueprints/admin.py
```diff
@@ -586,26 +626,31 @@ def post(self):
...
-        if errors:
-            return render_template(
-                self.template, errors=errors, messages=messages)
-        messages = {'Success': 'Permission created.'}
-        return redirect(url_for('admin.permissions',
-                                messages=messages))
+
+        # display any errors and redirect to the newly created permission.
+        self.flash_api_errors(errors)
+        flash('Success: Permission created.')
```
wholmgren, line 652: can we simultaneously have errors and success?

sfa_dash/templates/forms/admin/permission_macros.jinja
lboeman: Yeah, thanks this is probably what caused the issue to begin with.

sfa_dash/blueprints/admin.py
lboeman: Yes. So in reviewing this I realized the current behavior is really confusing and obfuscates what's actually happening.

Current behavior:
1. Try to create a permission, if it fails, show the form with errors.
2. If the permission was created successfully, try to add the selected objects to the permission. Collect error and success messages. If errors occurred, show the form and print all of the messages. This is deceptive because we created a permission, but we're still on the "create a new permission" form. The user has no indication that they can access the permission that was created and will probably just try to create another one.

Behavior introduced in this PR:
1. Same as above.
2. Try to add the objects to the new permission, collecting an error for any failure. Then redirect the user to the new permission page with error messages(if they exist). The user will be able to see the new permission metadata and the objects that were successfully added along with error messages about those that failed. 

Other than these bugs, there are only two conditions in which users should see these errors:
-They don't have create permissions. They should see a failure at step 1 above.
- They have a `create permissions` permission, but not `update all permissions`. Then, they could create the permission buy adding objects would fail. This might be confusing at first, but now that they land on the created permission page they will see a list of errors and won't see the "add objects to permission" button.

wholmgren: thanks that makes sense!

PR #340, opened 2020-07-16
Hide on create perm

### LINKED ISSUE #337: create observation, forecast, etc permissions should hide selection menu
When creating a new permission of type `create`, the object selection menu is not relevant and should be hidden like so:

Screen Shot 2020-07-16 at 12 05 23 PM

lboeman:
closes #337 

- Added a description section to explain why we're hiding the "applies to all" option and the listing when a user is creating a permission with create action. When the user selects 'create' the section expands, and the 'applies to all' and listing collapses. The opposite occurs when the user selects a different action.  Here's what I have for the description: 

> Create permissions don't require any further configuration, but creating new metadata may require other permissions. For example, creating an Observation at a Site requires Create permissions for Observations and Read permission on the Site.

- Hid the "Add objects to this permission" button, and the table of objects on permission pages when the action of that permission is 'create'.

### DIFF
--- a/sfa_dash/templates/forms/admin/permissions_form.html
+++ b/sfa_dash/templates/forms/admin/permissions_form.html
@@ -29,7 +29,7 @@
   <p>Select the action this permission will allow.</p>
   <input type="radio" name="permission-action" value="create" required>Create
   <input type="radio" name="permission-action" value="read">Read
-  {% if data_type not in ['fpermission-orecast', 'observation', 'site'] %}
+  {% if data_type not in ['forecast', 'observation', 'site'] %}
   <input type="radio" name="permission-action" value="update">Update
   {% endif %}
   <input type="radio" name="permission-action" value="delete">Delete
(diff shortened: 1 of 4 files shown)


wholmgren:
Description is good. Wondering if we can be more complete though. Do the create permissions for Site, Reports, or Aggregates require anything else? If not, perhaps the following for obs/fx/probfx and nothing for the others:

```
Create permissions don't require any further configuration, but creating new
{{ data_type }} requires <i>Read</i> permissions on the associated Site.
```

Was thinking about the grammar and realized that all buttons under the "Create a New Permission" are plural except for "Observation".

sfa_dash/templates/forms/admin/permissions_form.html
```diff
@@ -27,18 +27,33 @@
...
+  <input type="radio" name="permission-action" value="write_values">Write Values
+  <input type="radio" name="permission-action" value="delete_values">Delete Values
   {% endif %}
 </div>
+<div id="create-permission-explanation" class="collapse">
+{% set perm_requirements = {
+    'site': '',
+    'observation': '<i>Read</i> permissions on the associated Site',
+    'forecast': '<i>Read</i> permissions on the associated Site',
+    'aggregate': '<i>Read</i> and <i>Read Values</i> permissions on Observations to make up the aggregate',
```
wholmgren, line 47:
```suggestion
    'aggregate': '<i>Read</i> and <i>Read Values</i> permissions on Observations that make up the Aggregate',
```

PR #349, opened 2020-09-01
Reference forecast inclusion options

lboeman:
Adds options to include reference forecasts as their own forecast, observation pair or to use them only for the skill metric as discussed in https://github.com/SolarArbiter/solarforecastarbiter-core/issues/511. Controlled by a radio button selection just above the selection box. 
A screenshot is included below.
Screenshot from 2020-09-01 16-29-01

Also fixes an issue with retaining reference forecast selection for probabilistic forecasts.

### DIFF
--- a/sfa_dash/static/js/report-form-handling.js
+++ b/sfa_dash/static/js/report-form-handling.js
@@ -29,35 +29,37 @@ function addPair(
      *  @param {string} forecast_type
      *      The type of forecast in the pair.
      */
-
-    var new_object_pair = $(`<div class="pair-container object-pair object-pair-${pair_index}">
-            <div class="input-wrapper">
-              <div class="col-md-12">
-                <div class="object-pair-label forecast-name-${pair_index}"><b>Forecast: </b>${fx_name}</div>
-                <input type="hidden" class="form-control forecast-value" name="forecast-id-${pair_index}" required value="${fx_id}"/>
-                <div class="object-pair-label truth-name-${pair_index}"><b>Observation: </b> ${truth_name}</div>
-                <input type="hidden" class="form-control truth-value" name="truth-id-${pair_index}" required value="${truth_id}"/>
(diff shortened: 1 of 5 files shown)


wholmgren:
That looks great. Minor things:

At the risk of being too long, what about "Also include as standard forecast/observation pair." Or even "Also include as standard forecast/observation pair with full metrics." 

The search box could be bigger so that the full hint is shown.

Is the "binary forecasts" text still only on the probabilistic form? It wouldn't make sense on the other forms.

lboeman:
> Is the "binary forecasts" text still only on the probabilistic form? It wouldn't make sense on the other forms.

Yes this is still only on the probabilistic form, but my understanding is that it was meant to clarify the skill metric being calculated against a matching constant value in the reference forecast?

lboeman:
Here's an updated screenshot with the longer text. I think it works well. It does wrap on small screens, but I think it's worth the tradeoff.
Screenshot from 2020-09-02 09-51-24

sfa_dash/static/js/report-form-handling.js
```diff
@@ -550,6 +568,20 @@ function createPairSelector(){
                            "Unset",
                            null,
             );
+            if ($('[name="include-reference"]:checked').val() == "true"
+                && ref_id != null){
+                console.log
```
alorenzo175, line 573:
```suggestion
```

lboeman:
> The layout looks good. I couldn't tell by skimming the javascript, but are duplicate pairs of reference fx/obs are taken care of somewhere?

No, that's a good point. I'll add handling for that.

wholmgren:
good to merge and test in dev?

lboeman:
I've added a global `object_pairs` array helper functions `report_utils.try_insert_pair` and `report_utils.remove_pair` to keep track of the pairs added to the form. `try_insert_pair` returns a boolean based on if the pair already exists, and is used as the conditional statement controlling whether or not we add an object to the DOM when the user clicks "add obs/fx pair".

PR #350, opened 2020-09-02
Minor issue cleanup

lboeman:
- Casts form fields to integers when necessary during form parsing or api to form conversion to avoid error thrown when string representation float e.g. "1.0" cannot be cast to an int.
- Adds missing `required` attribute to `constant-values` field of cdf forecast forms.
- Always show modeling parameters on site form, disable/enable based on value of the "Site type" field.

### DIFF
--- a/sfa_dash/templates/forms/base/base_report_form.html
+++ b/sfa_dash/templates/forms/base/base_report_form.html
@@ -18,7 +18,7 @@
 <form action="{{ url_for('forms.create_'+report_type+'_report') }}" method="post" id="report-form" onSubmit="return report_utils.validateReport();">
   <div class="form-group">
      <div class="form-element full-width">
-       <label for="start">Name:</label><br/>
+       <label for="name">Name:</label><br/>
        <div class="input-wrapper">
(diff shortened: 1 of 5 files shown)


sfa_dash/templates/forms/base/base_report_form.html
```diff
@@ -18,6 +18,7 @@
 <form action="{{ url_for('forms.create_'+report_type+'_report') }}" method="post" id="report-form" onSubmit="return report_utils.validateReport();">
   <div class="form-group">
      <div class="form-element full-width">
+       <label for="start">Name:</label><br/>
```
alorenzo175, line 21: shouldn't the label be for "name"? Although now I see no ids for the inputs, so maybe these need a little more cleanup

alorenzo175 (approved):

sfa_dash/templates/forms/base/base_report_form.html
lboeman: This is going to be kind of a big job to clean up, with a lot of mislabeling happening in javascript. I've opened up #352 as future work.

PR #351, opened 2020-09-03
Probabilistic forecast group plots

lboeman:
Adds probabilistic forecast plots to the cdf forecast group pages.

### DIFF
--- a/sfa_dash/api_interface/cdf_forecast_groups.py
+++ b/sfa_dash/api_interface/cdf_forecast_groups.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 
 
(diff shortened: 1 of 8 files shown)


lboeman:
Added probabilistic forecast plots for forecasts with axis X and Y. I've reused explanatory text for axis = Y plots from reports. The legend was also removed for axis = Y, due to a quirk where hiding any of the traces causes plotly to redraw the color fill to the next line, misrepresenting the data. Neither axis type pages have a download button, and instead have a note that data can be downloaded from the individual constant value forecasts found in the table.

Screenshot of an axis=y forecast :
y_axis_prob_timeseries

Screenshot of an axis=x forecast :
x_axis_prob_timeseries

Test failures are currently due to dependency on https://github.com/SolarArbiter/solarforecastarbiter-core/pull/551

lboeman:
Included better labeling used in reports for the legend/hover of x axis plots and the hover tool of the y axis plots.
y axis:
y_axis_prob_timeseries
x axis:
x_axis_prob_timeseries

wholmgren (approved):
@alorenzo175 ?

sfa_dash/api_interface/cdf_forecast_groups.py
```diff
@@ -1,14 +1,36 @@
...
+    Returns
+    -------
+    pandas.DataFrame
+        Dataframe where column names are constant values.
+    """
+    constant_values = metadata['constant_values']
+    data = pd.DataFrame()
+    for cv in constant_values:
+        req = get_cdf_values(cv["forecast_id"], **kwargs)
+        data[str(cv['constant_value'])] = json_payload_to_forecast_series(req)
```
alorenzo175, line 32: I wonder about replacing this with ``concurrent.futures.ThreadPoolExecutor.map`` or similar since it will mostly be waiting for the api to respond and there might be a bunch of constant valvues

alorenzo175: I don't recall what goes into get_request or if threading will be a problem for the session

alorenzo175 (commented):
Looks good. Just think there might be a easy, non-negligible speed up when getting the data

sfa_dash/api_interface/cdf_forecast_groups.py
lboeman: Alright, I got this with a wrapper function that copies the current app context, and using flasks `copy_current_request_context` wrapper. Exceptions should be reraised when trying to access responses from the iterator.

alorenzo175: Cool! sorry, didn't think about the flask request context for the oauth session etc.

Will something like this work to continue processing while some threads are waiting for responses?
```python
def request_constant_value_values(cv):
    return cv['constant_value'], json_payload_to_forecast_series(get_cdf_values(cv["forecast_id"], **kwargs))

all_series = ...
return pd.DataFrame.from_items(all_series)
```

lboeman: I'm not sure if what I've done in this update will have the intended affect. I think what you're going for is to pass the generator before the futures are complete? I'm not super familiar with concurrent.futures, but I did see that something I missed in my first stab at this was shutting down the executor, so I moved things into a with statement so we don't leak threads.

alorenzo175: I think it does what I was getting at, making the request and processing the json into a series in each thread as the data comes in, then gather all the series to make the dataframe

alorenzo175 (approved):

PR #355, opened 2020-09-15
add support for seconds in report start/end datetimes closes #353

### LINKED ISSUE #353: fix ISO 8601 parsing in Create New Report
Screen Shot 2020-09-09 at 2 37 21 PM

Both of these are valid ISO 8601. And the "end" is the same format as the form is prepopulated with when clicking "clone report".

I think it's just a matter of allowing (but not requiring) seconds in the parser

lboeman:


alorenzo175 (approved):

lboeman:
Working on an update for better regex here.

alorenzo175:
still requiring the T and only UTC? What do you think of copying the marshmallow regex and removing the labels?

alorenzo175:
good whenever you are

PR #356, opened 2020-09-15
Add reference forecast requirement message for QSS and BSS

lboeman:
Adds the same warning message used on the deterministic form for the Skill metric to BSS and QSS on the probabilistic report form.

Screenshot from 2020-09-15 14-48-59


wholmgren (approved):
Looks good. Are we still keeping log messages?

lboeman:
> Looks good. Are we still keeping log messages?

Nope, thanks.

PR #362, opened 2020-10-02
add delay to execution of table search functions

### LINKED ISSUE #359: Add delay to search functions
Currently fires on keyup, this is extremely lag-inducing for the listing pages as the number of objects has grown. Should be updated to fire after the user hasn't typed for some time (maybe 1seond).

lboeman:
closes #359 
Adds a 300ms delay before firing the search functions found on  listing pages (e.g. `/forecasts/single`). Only allows one call to be queued at a time. As a user types, keyups queue a new search function call with the current contents of the search box and cancel any pending call. When the 300ms delay expires, the Search function is called and filters the listing.

### DIFF
--- a/sfa_dash/static/js/table-search.js
+++ b/sfa_dash/static/js/table-search.js
@@ -14,6 +14,26 @@
  * https://codepen.io/adobewordpress/pen/gbewLV
  */
 $(document).ready(function() {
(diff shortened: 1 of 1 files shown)


wholmgren (approved):
assuming it works...

PR #363, opened 2020-10-02
Set page titles

### LINKED ISSUE #202: make page titles more descriptive
like we did in https://github.com/SolarArbiter/SolarArbiter.github.io/pull/130

lboeman:
Closes #202

### DIFF
--- a/sfa_dash/templates/dash/data.html
+++ b/sfa_dash/templates/dash/data.html
@@ -30,7 +30,12 @@
 
 {% if page_title is defined or breadcrumb is defined %}
 <h2 class="page-title">
-{{ page_title }}{{ breadcrumb|safe }}
+{# display a breadcrumb if present, otherwise use page_title #}
+{% if breadcrumb is defined %}
+{{ breadcrumb|safe }}
+{% else %}
+{{ page_title }}
+{% endif %}
 </h2>
 {% endif %}
 
--- a/sfa_dash/templates/head.html
+++ b/sfa_dash/templates/head.html
@@ -4,7 +4,17 @@
 <meta name="author" content="">
 <link rel="icon" href="/static/images/favicon.ico">
 
-<title>Solar Forecast Arbiter</title>
+<title>
+{% if metadata is defined and 'name' in metadata %}
+{{ metadata['name'] }}
+{% elif metadata is defined and 'report_parameters' in metadata %}
+{{ metadata['report_parameters']['name'] }}
+{% elif page_title is defined %}
+{{ page_title }}
+{% else %}
+Solar Forecast Arbiter Dashboard
+{% endif %}
+</title>
 
 <!-- Bootstrap core CSS -->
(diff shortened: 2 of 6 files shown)


sfa_dash/templates/head.html
```diff
@@ -4,7 +4,17 @@
 <meta name="author" content="">
 <link rel="icon" href="/static/images/favicon.ico">
 
-<title>Solar Forecast Arbiter</title>
+<title>
+{% if metadata is defined and 'name' in metadata %}
+{{ metadata['name'] }}
+{% elif metadata is defined and 'report_parameters' in metadata %}
+{{ metadata['report_parameters']['name'] }}
+{% elif page_title is defined %}
```
wholmgren, line 12: should this be the first `if` block?

sfa_dash/templates/head.html
```diff
@@ -4,7 +4,17 @@
 <meta name="author" content="">
 <link rel="icon" href="/static/images/favicon.ico">
 
-<title>Solar Forecast Arbiter</title>
+<title>
+{% if metadata is defined and 'name' in metadata %}
+{{ metadata['name'] }}
+{% elif metadata is defined and 'report_parameters' in metadata %}
+{{ metadata['report_parameters']['name'] }}
```
wholmgren, line 11: We use `page title | Solar Forecast Arbiter` on the informational website. I could go either way on the `| Solar Forecast Arbiter` here on the dashboard.

sfa_dash/templates/dash/data.html
```diff
@@ -30,7 +30,12 @@
 
 {% if page_title is defined or breadcrumb is defined %}
 <h2 class="page-title">
-{{ page_title }}{{ breadcrumb|safe }}
+{# display a breadcrumb if present, otherwise use page_title #}
+{% if breadcrumb is defined %}
+{{ breadcrumb|safe }}
+{% else %}
+{{ page_title }}
+{% endif %}
```
wholmgren, line 38: what does this address?

lboeman: `page_title` is used to set a static non-link heading on the page, and `breadcrumb` is expected to be the series of links separated with `/`. This is why the page title setting above first checks for a metadata object to pull a name from. The individual forecast/obs pages only have a breadcrumb.

wholmgren: ah, so this doesn't actually change anything or fix an issue, it's just more explicit about what's going on?

lboeman: To get the page `<title>` attributes working I added `page_title` to pages that were missing it, so separating was to explicitly ensure we don't print both.

lboeman: To answer your original question, This is to allow for providing a `page_title` and a `breadcrumb` without printing both at the top of the page. In that case, the `page_title` is only used in the `<title>` tag and the `breadcrumb` is displayed at the top of the page. For example, on the /Sites where we want a `page_title` of "Sites" and a breadcrumb like `/<a href="/sites">Sites</a>`.

wholmgren (approved):

PR #364, opened 2020-10-08
Deletion pairs

### LINKED ISSUE #234: deleting a report should show object pair names
maybe display the pairs like on the report creation page

Screen Shot 2020-04-23 at 4 06 54 PM

bonus points if the names are links too

_Originally posted by @wholmgren in https://github.com/SolarArbiter/solarforecastarbiter_dashboard/issues/232#issuecomment-618722976_

lboeman:
closes #234 
Adds object pairs with link names and cost parameters to the report metadata block of delete report pages. Names of each object are links to their dashboard page.
report_w_pairs_costs

### DIFF
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -249,40 +249,68 @@ class DeleteReportView(BaseView):
         forecast_type = pair['forecast_type']
 
         if forecast_type == 'forecast' or forecast_type == 'event_forecast':
-            forecast_metadata = forecasts.get_metadata(pair['forecast'])
+            try:
+                forecast_metadata = forecasts.get_metadata(pair['forecast'])
+            except DataRequestException:
+                forecast_metadata = None
+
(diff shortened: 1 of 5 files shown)


wholmgren (approved):
You even got the bonus points ⭐ !

Does this need any special cases for event or probabilistic forecasts?

Screen Shot 2020-10-09 at 3 23 58 PM

lboeman:
> You even got the bonus points  !
> 
> Does this need any special cases for event or probabilistic forecasts?
> 
> Screen Shot 2020-10-09 at 3 23 58 PM

Ah, I hadn't considered adding the constant values to names. I think adding the `Prob(...)` suffixes as they appear in reports should work? Nesting would be somewhat complicated, but is doable if that's a valuable addition.

Event forecasts are treated exactly as normal deterministic forecasts are.

wholmgren:
> I think adding the Prob(...) suffixes as they appear in reports should work? 

Fine with me

PR #365, opened 2020-10-09
Allow report viewing with partial permissions

### LINKED ISSUE #134: Should be able to view an incomplete report without 'read_values' permissions
We should be able to read the report metrics ('read' permission) and without reading the report values ('read_values' permission) on the dashboard. This probably requires changes to API and/or core as well.

lboeman:
closes #134 
Relies on https://github.com/SolarArbiter/solarforecastarbiter-api/pull/280.
Removes the timeseries/scatter plot specifications from the report template arguments if the report `values` field is empty or if the maximum number of data points to plot is exceeded. 

This also fixes a bug where if the maximum datapoints were exceeded the warning was not displayed. This was something missed in the migration to plotly. The warning message is now displayed at the top of the page in the message section, since Plotly does not have a `<div>` to override with a message like bokeh did.

### DIFF
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -160,24 +160,31 @@ class ReportView(BaseView):
             'report_template': report_template,
             'dash_url': request.url_root.rstrip('/'),
             'include_metrics_toc': False,
-            'includes_bokeh': True,
             'metadata': {
                 'report_parameters': self.metadata['report_parameters']},
         })
-        if not include_timeseries:
+        # If we're not including timeseries, or values could not be read,
+        # pop all of the timeseries/scatter spec template arguments
+        if not include_timeseries or not self.metadata['values']:
+            report_status = self.metadata.get('status')
             # display a message about omitting timeseries
             download_link = url_for('data_dashboard.download_report_html',
                                     uuid=self.metadata['report_id'])
-            script = ''
-            div = f"""<div class="alert alert-warning">
-    <strong>Warning</strong> To improve performance timeseries plots have been
-    omitted from this report. You may download a copy of this report with the
-    timeseries plots included:
-    <a href="{download_link}">Download HTML Report.</a></div>"""
-            report_kwargs.update({
-                'timeseries_div': div,
-                'timeseries_script': script,
-            })
+            report_kwargs.pop('timeseries_spec', None)
+            report_kwargs.pop('timeseries_prob_spec', None)
+            report_kwargs.pop('scatter_spec', None)
+            if not include_timeseries and report_status == 'success':
+                flash(
+                    f"""<strong>Warning</strong> To improve performance
+                    timeseries plots have been omitted from this report. You
+                    may download a copy of this report with the timeseries
+                    plots included:
+                    <a href="{download_link}">Download HTML Report.</a>""",
+                    'warning')
+            elif not self.metadata['values'] and report_status == 'success':
+                flash('Could not load report values. Timeseries and scatter '
+                      'plots will not be included.',
+                      'warning')
         self.template_args = report_kwargs
 
     def set_metadata(self, uuid):
--- a/sfa_dash/blueprints/reports.py
+++ b/sfa_dash/blueprints/reports.py
@@ -175,15 +175,18 @@ class ReportView(BaseView):
             report_kwargs.pop('scatter_spec', None)
             if not include_timeseries and report_status == 'complete':
                 flash(
-                    f"""<strong>Warning</strong> To improve performance
-                    timeseries plots have been omitted from this report. You
-                    may download a copy of this report with the timeseries
-                    plots included:
+                    f"""<strong>Warning</strong> Too many timeseries points
+                    detected. To improve performance time series plots have
+                    been omitted from this report. You may download a copy
+                    of this report with the timeseries plots included:
                     <a href="{download_link}">Download HTML Report.</a>""",
                     'warning')
             elif not self.metadata['values'] and report_status == 'complete':
-                flash('Could not load report values. Timeseries and scatter '
-                      'plots will not be included.',
+                flash('Could not load any time series values of observations '
+                      'or forecasts. Timeseries and scatter plots will not '
+                      'be included. You may require the `read_values` '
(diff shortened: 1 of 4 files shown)


sfa_dash/blueprints/reports.py
```diff
@@ -158,24 +158,31 @@ def set_template_args(self):
...
+            if not include_timeseries and report_status == 'success':
+                flash(
+                    f"""<strong>Warning</strong> To improve performance
+                    timeseries plots have been omitted from this report. You
+                    may download a copy of this report with the timeseries
+                    plots included:
+                    <a href="{download_link}">Download HTML Report.</a>""",
+                    'warning')
+            elif not self.metadata['values'] and report_status == 'complete':
+                flash('Could not load report values. Timeseries and scatter '
```
wholmgren, line 183: Users might be confused by the term "report values". We really mean the time series of observation and forecast values, right? If so, how about "Could not load time series of observations and forecasts. Time series and scatter plots will not be included."

wholmgren: Also this is only the case when *none* of the timeseries values are available, right? In that case maybe "Could not load any..."

sfa_dash/blueprints/reports.py
```diff
@@ -158,24 +158,31 @@ def set_template_args(self):
...
-                'timeseries_div': div,
-                'timeseries_script': script,
-            })
+            report_kwargs.pop('timeseries_spec', None)
+            report_kwargs.pop('timeseries_prob_spec', None)
+            report_kwargs.pop('scatter_spec', None)
+            if not include_timeseries and report_status == 'success':
+                flash(
+                    f"""<strong>Warning</strong> To improve performance
+                    timeseries plots have been omitted from this report. You
```
wholmgren, line 177: I checked and we use both timeseries and time series in the documentation, but I think two words is better

sfa_dash/blueprints/reports.py
```diff
@@ -158,24 +158,31 @@ def set_template_args(self):
...
+                    may download a copy of this report with the timeseries
+                    plots included:
+                    <a href="{download_link}">Download HTML Report.</a>""",
```
wholmgren, line 180: this has me wondering if we should also offer a download without the time series plots.

sfa_dash/blueprints/reports.py
wholmgren: I'd like to say something about the reason for this is that there are so many data points, but I'm not sure how to clearly and concisely communicate that. The best I have is "Too many time series points detected. To improve..."

sfa_dash/blueprints/reports.py
```diff
@@ -158,24 +158,31 @@ def set_template_args(self):
...
-    omitted from this report. You may download a copy of this report with the
-    timeseries plots included:
-    <a href="{download_link}">Download HTML Report.</a></div>"""
-            report_kwargs.update({
-                'timeseries_div': div,
-                'timeseries_script': script,
-            })
+            report_kwargs.pop('timeseries_spec', None)
+            report_kwargs.pop('timeseries_prob_spec', None)
+            report_kwargs.pop('scatter_spec', None)
```
wholmgren, line 173: This introduces a coupling of the core and dashboard design, so we should add a comment to the relevant section of core. Something like "# if adding new plots, update sfa_dash/blueprints/reports.py to exclude when data does not exist" here: https://github.com/SolarArbiter/solarforecastarbiter-core/blob/7a6feca43c9eb6c90e72de892a5384575aa90c64/solarforecastarbiter/reports/template.py#L85-L97

sfa_dash/blueprints/reports.py
```diff
@@ -158,24 +158,31 @@ def set_template_args(self):
...
+            if not include_timeseries and report_status == 'success':
+                flash(
+                    f"""<strong>Warning</strong> To improve performance
+                    timeseries plots have been omitted from this report. You
+                    may download a copy of this report with the timeseries
+                    plots included:
+                    <a href="{download_link}">Download HTML Report.</a>""",
+                    'warning')
+            elif not self.metadata['values'] and report_status == 'complete':
```
wholmgren, line 182: is `report_status` documented somewhere? I couldn't tell you the difference between 'success' and 'complete'.

lboeman: This was an error on my part, "success" is not a valid report status. Opened an api issue to address this.

sfa_dash/blueprints/reports.py
lboeman: You're right, this is only the case when no timeseries values could be loaded, but could stem from two causes: lack of `read_values` permission on the report or a lack of `read_values` permissions on *all* of the included observations or forecasts, so I've added `You may require the 'read_values' permission on this report, or the included forecasts and observations.` to the warning, so that users know what they need to move forward.

sfa_dash/blueprints/reports.py
lboeman: Actually, looking into this I've realized we can remove this coupling. Since we're not actually interacting with the template anymore in the dashboard, and core checks if each "spec" is defined before including the plot or explanatory text, we can just adjust the core code to not define those values instead of the current setting them to `{}` and we won't have to intervene here.

lboeman:
I've removed the handing of those template arguments here, instead only warning the user about possible permissions issues. I've opened https://github.com/SolarArbiter/solarforecastarbiter-core/pull/585 to further address this issue since there were some cases where the report rendering code would choke due to some missing data.

sfa_dash/blueprints/reports.py
```diff
@@ -158,24 +158,31 @@ def set_template_args(self):
...
             'dash_url': request.url_root.rstrip('/'),
             'include_metrics_toc': False,
-            'includes_bokeh': True,
             'metadata': {
                 'report_parameters': self.metadata['report_parameters']},
         })
-        if not include_timeseries:
-            # display a message about omitting timeseries
+        # If we're not including timeseries, or values could not be read,
+        # pop all of the timeseries/scatter spec template arguments
```
wholmgren, line 165: comment looks like it's no longer relevant

wholmgren (approved):

PR #376, opened 2020-10-22
compress zipped files

### LINKED ISSUE #324: compress zip files
add https://docs.python.org/3/library/zipfile.html#zipfile.ZIP_DEFLATED to check_sign_zip

alorenzo175:
closes #324


lboeman (approved):
Looks good and works great. Creating a report with the test ghi forecast/observation ends up with a 28.6kB zip file and 278.7kB HTML file.

PR #379, opened 2020-10-23
Delete obs from aggregate

lboeman:
Adds a delete button to the aggregate observation table if the user has `update` permission.
Screenshot from 2020-10-23 16-52-43
The deletion confirmation page displays a table of the times the observation has been added to the aggregate, and warns the user that they are removing all of them.
Screenshot from 2020-10-23 16-55-08


lboeman:
Updated the observation table to only display observation names, organizations and delete buttons once to make workflow more apparent. 
Screenshot from 2020-10-26 15-54-27

alorenzo175 (approved):

PR #380, opened 2020-10-27
CRPSS

lboeman:
Adds CRPSS to the reference dependent metric listing so that a warning appears when no reference forecast is selected. Also fixes issue with multiple warnings. This should be all that's needed when CRPSS is added in core, as the list of metrics is built from `solarforecastarbiter.datamodel.ALLOWED_PROBABILISTIC_METRICS`.


wholmgren (approved):
looks good... assuming it works...

lboeman:
I intercepting and inserting the crpss metric in the `formdata_to_api_payload`, since those metrics are disabled because we're automatically controlling them. Should be good to go now.

PR #384, opened 2020-11-12
Download metadata

### LINKED ISSUE #298: add download buttons for metadata
Stakeholder suggestion. Would allow people to spec forecast requirements using the Arbiter's forms and then send to other people that are not already using the Arbiter.

lboeman:
closes #298 

Adds a **Download Metadata** button to the top-right corner of the metadata section for sites, forecasts, observations and aggregates. These downloads do not include `extra_parameters` as there are some issues that arise from parsing the metadata from jinja. If we want to include them, endpoints to fetch the metadata and return a  file could be implemented.

Screenshot from 2020-11-12 12-12-09

Built on top of #383, will need to be rebased.

### DIFF
--- a/sfa_dash/static/css/styles.css
+++ b/sfa_dash/static/css/styles.css
@@ -1,255 +1,305 @@
 /* Bootstrap overrides */
-@media (min-width: 768px){
-  .fixed-top{
+@media (min-width: 768px) {
+  .fixed-top {
     position: fixed;
   }
 }
-@media (max-width: 767px){
-  .fixed-top{
+
+@media (max-width: 767px) {
+  .fixed-top {
     position: relative;
   }
 }
-.fixed-top{
+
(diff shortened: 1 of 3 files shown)


sfa_dash/static/css/styles.css
```diff
@@ -148,6 +148,7 @@ header nav{
 	width: inherit;
 }
 .data-metadata{
+    position:relative;
```
wholmgren, line 151: indent and space after colon?

wholmgren (approved):
We can go ahead with this but let's make a 1.x issue for including the extra parameters in the download.

sfa_dash/static/css/styles.css
lboeman: Just went ahead and ran a linter on styles.css to fix lots of formatting issues.

PR #385, opened 2020-11-12
Move upload form to obs/fx page

### LINKED ISSUE #196: Move upload data controls to obs/fx/cdf pages
After #184, it was brought up that having a separate page for uploads is awkward. We should move the upload ability to the same pages, and ideally provide a sample document for format reference and return the user to view the data they just uploaded.

lboeman:
closes #196 
Makes the **Upload Data** button control the collapse of the upload data form. I've left the button just below the metadata section, and when clicked the upload form expands below. Screenshots below.

Expanded:
expanded

After errors:
errors

### DIFF
--- a/sfa_dash/blueprints/form.py
+++ b/sfa_dash/blueprints/form.py
@@ -1,5 +1,7 @@
 import json
 from json.decoder import JSONDecodeError
+from requests.exceptions import HTTPError
+
 
 from flask import (Blueprint, render_template, request,
                    abort, redirect, url_for, flash)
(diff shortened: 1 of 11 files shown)


lboeman:
After some discussion landed on a design more in line with the proposal in the issue. There is an **Upload Data** button for expanding the section at the bottom of the page. Screenshot of the final product:
upload section

wholmgren (approved):
The code looks correct to me, but do we need to manually check that it works on all types of `data_type`? I forgot about all of the switches behind the scenes.

lboeman:
> The code looks correct to me, but do we need to manually check that it works on all types of `data_type`? I forgot about all of the switches behind the scenes.

Locally I've checked that this works with and without bad data for observations forecast and probabilistic forecast constant values. I did discover an API bug where empty data causes a 500, so I'll open an issue/pr there.

PR #387, opened 2020-11-12
redirect to location view after deletion

### LINKED ISSUE #103: redirect after delete
Deleting an observation sends me to `/observations`. This surprised me because I arrived at the observation by going through a site and its observation listing. I would have preferred to go back to the site observation listing. Is there a way to account for how the user arrived to the page when deciding where to redirect to?

lboeman:
Closes #103 
When deleting an observation, forecast or probabilistic forecast, users will now be redirected to the associated site/aggregate's listing for that type.

### DIFF
--- a/sfa_dash/blueprints/delete.py
+++ b/sfa_dash/blueprints/delete.py
@@ -68,7 +68,25 @@ class DeleteConfirmation(DataDashView):
             # the confirmation page, redirect to confirm.
             return redirect(confirmation_url)
         try:
+            self.metadata = self.api_handle.get_metadata(uuid)
+        except DataRequestException:
+            redirect_url = url_for(f'data_dashboard.{self.data_type}s')
+        else:
+            if self.metadata.get('site_id') is not None:
+                if self.data_type == 'site':
+                    redirect_url = url_for('data_dashboard.sites')
+                else:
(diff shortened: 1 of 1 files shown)


sfa_dash/blueprints/delete.py
```diff
@@ -67,8 +67,26 @@ def post(self, uuid):
...
+        except DataRequestException:
+            redirect_url = url_for(f'data_dashboard.{self.data_type}s')
+        else:
+            if self.metadata.get('site_id') is not None:
+                if self.data_type == 'site':
+                    redirect_url = url_for('data_dashboard.sites')
+                else:
+                    redirect_url = url_for(f'data_dashboard.{self.data_type}s',
+                                           site_id=self.metadata['site_id'])
+            else:
```
alorenzo175, line 81: only data_types that definitely have site_id or aggregate_id will get here? worth making this `elif 'aggregate_id' in self.metadata` and making `else` just the old redirect_url ?

alorenzo175 (approved):

sfa_dash/blueprints/delete.py
lboeman: Good call, thanks.

PR #402, opened 2021-01-08
Handle Download Data Error

lboeman:
https://sentry.io/organizations/solararbiter-rx/issues/2138278848/?project=1472478&query=url%3A%22https%3A%2F%2Fdashboard.solarforecastarbiter.org%2Fobservations%2F063d42da-99fe-11e9-8674-0a580a8200c9%22

Handles two problems from the above Sentry error:
1. If the download failed, `set_template_args` wasn't called so an exception was raised trying to access the `template_args` attribute. Instead of handling this exception inside of the `download_timeseries` helper the View classes that utilize it handle the exception, flash the error and return their get method.
2. Raise an error when a NaT type makes it into the form. In the sentry error above, query parameters of `start=NaT&end=NaT` were passed. This should be handled by the `parse_start_end_from_querystring` method of `BaseView`, which should populate the start and end inputs on the page with the most recent valid times.  It appears there may be a case where the javascript to setup the time parameters is not running properly on IE11. More investigation is needed there. For the time being, I've adjusted the `format_download_params` function to raise an error if a NaT value gets through, and send the exception to Sentry for more context, The use will then be presented with an "invalid datetime" error instead of the 500 with the offending query parameters removed.

### DIFF
--- a/sfa_dash/blueprints/aggregates.py
+++ b/sfa_dash/blueprints/aggregates.py
@@ -385,4 +385,10 @@ class AggregateView(BaseView):
     def post(self, uuid):
         """Download endpoint.
         """
-        return download_timeseries(self, uuid)
+        try:
+            data_response = download_timeseries(self, uuid)
+        except DataRequestException as e:
+            self.flash_api_errors(e)
+            return self.get(uuid)
+        else:
+            return data_response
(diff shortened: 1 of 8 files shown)


lboeman:
Confirmed that IE11 issues are due to unsupported javascript. Edge seems to handle the dashboard just fine. I've added a Warning banner when Internet explorer is detected to warn users that the dashboard will not function properly. 
Screenshot from 2021-01-08 12-02-25

Microsoft is ending support for IE11 in it's own apps this year, so it's a pretty safe bet not to support it for the Arbiter. https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666

wholmgren (approved):

PR #415, opened 2021-05-10
Improve Datetime widgets

### LINKED ISSUE #141: good, consistent datetime selection interfaces
We need good, consistent date selection interfaces across the dashboard. @cwhanse commented that the Observation/Forecast date selection box requires too much clicking and it would be more convenient to type in the fields. I'm sure that some other users would find that typing in fields is annoying (I think it's annoying for reports).

lboeman:
closes #141
Adopts date time input style from [Solar Performance Insight](https://github.com/SolarPerformanceInsight/solarperformanceinsight). Still needs to be adapted to the report form.

### DIFF
--- a/LICENSES/SOLARPERFORMANCEINSIGHT_LICENSE
+++ b/LICENSES/SOLARPERFORMANCEINSIGHT_LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 SolarPerformanceInsight
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


lboeman:
This should be working for data pages and the report creation page now. Interface looks as shown below.
Screenshot_2021-05-11 GHI Instrument 1

wholmgren:
Thanks. Needs a copy of the SPI license. Probably follow the pattern in [core](https://github.com/SolarArbiter/solarforecastarbiter-core/tree/master/LICENSES). FYI @cwhanse 

Is this a test in dev kind of thing? Or is @alorenzo175 set up to test?

wholmgren (approved):

PR #418, opened 2021-05-12
Pin Werkzeug version.

lboeman:
Workzeug 2.0.0 triggered https://sentry.io/organizations/solararbiter-rx/issues/2394111331/?project=1472478

Pins the werkzeug version to avoid these issues until we can address updating

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 Flask==1.0.2
+jinja2==2.10
 Flask-SeaSurf==0.2.2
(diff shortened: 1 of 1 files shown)


wholmgren:
should we pin other requirements in this PR or leave that for another PR?

lboeman:
> should we pin other requirements in this PR or leave that for another PR?

I think leave for another PR that perhaps updates all dependencies and then freezes versions or locks the current versions in place?

PR #424, opened 2021-07-21
Make report header rules more specific

### LINKED ISSUE #422: Report timeseries plots displaying at very small height
Height of report timeseries plots is displaying only enough to show the top item of the legend. This may be a css, or plotly version issue.

lboeman:
closes #422 
The selector for introducing pseudo elements to report headers such that they clear the sticky site header was too generic and effecting plotly's ability to layout correctly. This updates those selectors to only hit the headers on the page, resulting in correct plotly height layouts.


wholmgren (approved):
good catch

PR #425, opened 2021-07-21
Support Report Timezone

### LINKED ISSUE #423: Add timezone selection to report form
Add ability to select timezone when defining report parameters. Should inform the user that timezone will be inferred if not selected.

lboeman:
closes #423 
Adds the option for a user to supply a timezone for a report, with the default option being "Infer from selections. I've dropped it in right above the start/end inputs like so: 
Screenshot from 2021-07-21 14-44-07

### DIFF
--- a/sfa_dash/form_utils/tests/test_converter.py
+++ b/sfa_dash/form_utils/tests/test_converter.py
@@ -629,3 +629,49 @@ def test_report_converter_apply_crps_crpss(pair, expected_metrics):
     updated = converters.ReportConverter.apply_crps(params)
     assert updated['metrics'] == expected_metrics
 
+
+def test_report_converter_payload_to_formdata_timezone_default(report):
+    form_data = converters.ReportConverter.payload_to_formdata(report)
+    assert form_data['report_parameters']['timezone'] == ""
+
+
+def test_report_converter_payload_to_formdata_timezone_null(report):
+    the_report = deepcopy(report)
+    the_report['report_parameters']['timezone'] = None
+    form_data = converters.ReportConverter.payload_to_formdata(the_report)
+    assert form_data['report_parameters']['timezone'] == ""
+
+
+@pytest.mark.parametrize("tz,expected", [
+    ("", None),
+    ("America/Pheonix", "America/Pheonix")
+])
+def test_report_converter_formdata_to_payload(report, tz, expected):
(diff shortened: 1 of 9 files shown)


sfa_dash/form_utils/tests/test_converter.py
```diff
@@ -629,3 +629,49 @@ def test_report_converter_apply_crps_crpss(pair, expected_metrics):
...
+def test_report_converter_payload_to_formdata_timezone_null(report):
+    the_report = deepcopy(report)
+    the_report['report_parameters']['timezone'] = None
+    form_data = converters.ReportConverter.payload_to_formdata(the_report)
+    assert form_data['report_parameters']['timezone'] == ""
+
+
+@pytest.mark.parametrize("tz,expected", [
+    ("", None),
+    ("America/Pheonix", "America/Pheonix")
```
wholmgren, line 647: is this intentionally misspelled? (Phoenix)

wholmgren (commented):
A little weird to specify a timezone but then be forced to input start/end in UTC. Can we do better with reasonable effort?

lboeman:
> A little weird to specify a timezone but then be forced to input start/end in UTC. Can we do better with reasonable effort?

Yes. I'll update that.

sfa_dash/form_utils/tests/test_converter.py
lboeman: hooked on feeownix

lboeman:
I've updated the javascript to keep the timezone labels on the report form in sync with the selected timezone.
When a user changes the timezone, the date/time stays the same because it seemed to me that it would be more common to enter times and then realize you forgot to change the timezone and not want to reenter dates. But that could be adjusted if there is utility in adjusting time appropriately as the timezone is altered.

I also have the code parsing the report's start/end and then localizing so that if we update timezones on any old reports, (where start/end are in UTC) cloning report parameters gives you the correct times.

lboeman:
@wholmgren I did confirm that current behavior is as we discussed. The stored value returned from the API is first localized to UTC and then adjusted to a timezone IF one exists.

Also updated the `/` to `-` and spaces after start. It seems like it would take more time than I'd like to spend to implement 0 padded numbers here, so we can save it for future work if needed.

wholmgren (approved):

PR #428, opened 2021-07-29
Add warnings about required reference forecasts to all skill score metrics

lboeman:
Imports the REQ_REF_FX variables from core to keep in sync.


sfa_dash/template_globals.py
```diff
@@ -15,6 +15,9 @@
     ALLOWED_VARIABLES,
     COMMON_NAMES,
 )
+from solarforecastarbiter.metrics.deterministic import _REQ_REF_FX as DETREF
+from solarforecastarbiter.metrics.probabilistic import _REQ_REF_FX as PROBREF
```
wholmgren, line 19: The event metrics module also contains this list. Even though it's currently empty I think it might be a good idea to include it here for the future. Always dangerous to import private variables across projects but it's probably ok here.

wholmgren (approved):

PR #433, opened 2021-08-13
Add buttons for cloning sites/obs/fx

### LINKED ISSUE #432: populate Site, Forecast, Observation forms with json uploads
I'd like to give users a way to easily clone sites, forecasts, and observations. This would allow users to experiment in their own user space.

I think a clone button could be implemented for sites without too much effort. It would be more difficult to do so for forecasts/observations because the user needs the create forecast/observation permission at the particular site. While that might not be so hard to check for, it might make for a fairly counterintuitive user interface.

An alternative is a workflow in which users can download json metadata, click create site/forecast/observation, and click upload json to populate the form. For forecasts and observations, the form needs only a fraction of the metadata in the downloaded json.

lboeman:
related to #432 
Took a second to flesh this out because it was mostly done. I've added clone buttons to sites, observations, forecasts and cdf forecasts. Cloning aggregates is complicated so that option is not implemented.

Just checks if the 'aggregate' or 'site' key is loaded into metadata for obs/fx before allowing cloning, so we can ensure that the user has access to that object. 

Just adds a button to the row of buttons below the metadata section on any given page.

Screenshot from 2021-08-13 11-09-43

### DIFF
--- a/sfa_dash/templates/data/site.html
+++ b/sfa_dash/templates/data/site.html
@@ -15,8 +15,8 @@
   <a rold="button" class="btn btn-primary btn-sm" href="{{ url_for('forms.update_site', uuid=site_id) }}">Update Metadata</a>
   {% endif %}
   {% if can_create('sites') %}
-<a role="button" class="btn btn-primary btn-sm" href="{{url_for('forms.clone_site', uuid=site_id)}}">Clone Metadata</a>
-{% endif %}
+  <a role="button" class="btn btn-primary btn-sm" href="{{url_for('forms.clone_site', uuid=site_id)}}">Clone Metadata</a>
+  {% endif %}
   {% if is_allowed('delete') %}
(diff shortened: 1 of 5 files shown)


sfa_dash/templates/data/site.html
```diff
@@ -14,6 +14,9 @@
   {% if is_allowed('update') %}
   <a rold="button" class="btn btn-primary btn-sm" href="{{ url_for('forms.update_site', uuid=site_id) }}">Update Metadata</a>
   {% endif %}
+  {% if can_create('sites') %}
+<a role="button" class="btn btn-primary btn-sm" href="{{url_for('forms.clone_site', uuid=site_id)}}">Clone Metadata</a>
+{% endif %}
```
wholmgren, line 19:
```suggestion
  <a role="button" class="btn btn-primary btn-sm" href="{{url_for('forms.clone_site', uuid=site_id)}}">Clone Metadata</a>
  {% endif %}
```

wholmgren (approved):
not an expert in this code but seems reasonable

PR #441, opened 2021-09-15
Add button padding

### LINKED ISSUE #436: declutter buttons
The new clone metadata button has pushed a couple of buttons into another row on my browser (maybe other window sizes already experienced this issue). Maybe a little bit of vertical pad?

Screen Shot 2021-08-19 at 8 28 04 AM

lboeman:
closes #436 
Adds some vertical margin to wrapped buttons.
Screenshot from 2021-09-15 14-22-12

### DIFF
--- a/sfa_dash/static/css/styles.css
+++ b/sfa_dash/static/css/styles.css
@@ -886,3 +886,7 @@ table.report-form-quality-flag-table tr td:nth-child(3),
 table.report-form-quality-flag-table tr th:nth-child(3) {
     text-align: center;
 }
+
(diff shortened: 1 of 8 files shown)


wholmgren (approved):
My only concern is if we've forgotten any templates that I'm not familiar with

lboeman:
> My only concern is if we've forgotten any templates that I'm not familiar with

Good call, even though some of these only have two buttons max, I went ahead and wrapped all of the buttons that appear in the same region of the page so that they'll wrap if we ever add more.

PR #450, opened 2021-10-04
Remove sidebar and adjust dashboard layout

### LINKED ISSUE #449: Remove sidebar in favor of flat navbar
In the interest of creating more horizontal space to help with text crowding on report tables etc. 

We should create a flat navbar across the bottom of the header. It would be nice if the site title etc would collapse away to avoid adding vertical space.

lboeman:
closes #449

Makes some changes to improve readability and layout of the dashboard.
- Reduces global font size
- Remove sidebar in favor of having the sites/aggregates/reports/fx and obs buttons up in the header.
- Adds a "Provider" column to reports.

reports_no_sidebar


lboeman:
@wholmgren I styled the nav at the top of the page a bit. You can see the style of hovered menu item on the left. If you think it's still a little tame for a navigation, I can flip the effect and make them dark buttons to stand out. thoughts?
Screenshot from 2021-10-05 10-20-36

wholmgren:
Maybe it would be more obvious if "Solar Forecast Arbiter" wasn't drawing so much visual attention. Can you try putting it on the right side or removing it entirely? I guess then it's not obvious how to return "home" but it's not like home is that useful (separate issue: It would be cool to have a "favorites" list or lists that showed up when you logged into the page). Maybe a small logo would work instead, but alas.

What about copying the github style? Bold and underline the div of the current section.

Screen Shot 2021-10-05 at 11 17 45 AM

lboeman:
What about this potential header style that steals the logo used in the API documentation? I've implemented the github-esque underling but left the hover effect.
Screenshot from 2021-10-05 12-08-52

What are you thinking in terms of a "favorites" menu? Allow users to favorite pages to quickly return to?

